PR #38, opened 2017-05-01
Grid view fix

cboddy:


ianopolous:
It was super painful to rebase off master because of the whitespace changes. Let's keep those to their own PRs in future, @cboddy .

PR #53, opened 2017-12-14
Implement a hex viewer as a default file viewer, and an ascii text viewer

ianopolous:


ianopolous:
peergosd-hexview

ianopolous:
peergos-hexviewer

ianopolous:
peergos-text-viewer

ianopolous:
I've added a basic text viewer too

PR #61, opened 2018-02-16
user errors in paste action are now shown

kevodwyer:


src/components/filesystem/index.js
```diff
@@ -400,8 +400,7 @@ module.exports = {
                             that.showSpinner = false;
 			});
                     }).exceptionally(function(throwable) {
-			progress.show = false;
-			that.errorTitle = 'Error moving file: ' + name;
+			that.errorTitle = 'Error moving file';
```
ianopolous: what was the reason for removing the name from the error?

kevodwyer: because name is not defined and the errorBody (line below) contains the filename

PR #64, opened 2018-02-21
minimal close button for open modal

kevodwyer:


ianopolous:
I think with this change we can also make the modal fullscreen. No reason for the border anymore.

PR #66, opened 2018-06-27
spinner text upon login/sign-up

kevodwyer:


ianopolous:
The text isn't centred in firefox:
screenshot-2018-6-28 peergos

PR #67, opened 2018-08-15
re-work the spinner message component

kevodwyer:


src/components/login/index.js
```diff
@@ -88,7 +88,7 @@ module.exports = {
             const that = this;
             this.showSpinner = true;
             return peergos.shared.user.UserContext.signIn(
-                    that.username, that.password, that.network, that.crypto, {"accept_0" : x => that.spinnerMessage = x}).thenApply(function(context) {
+                    that.username, that.password, that.network, that.crypto, {"accept" : x => that.spinnerMessage = x}).thenApply(function(context) {
```
ianopolous: I'm surprised this works. I thought we used _0 because that's what javac expects?

PR #69, opened 2019-01-16
Feat/granting write access

kevodwyer:


ianopolous:
@kevodwyer can you update the Peergos dependency? Then this can be merged.

PR #71, opened 2019-02-27
Fix file uploads and downloads for file sizes > 2 GB

ianopolous:


src/components/filesystem/index.js
```diff
@@ -550,11 +550,19 @@ module.exports = {
             this.changePath(this.getPath() + name);
         },
 
+	// This will only work up to a file size of 2^52 bytes (the biggest integer you can fit in a double)
+	// But who ever needed a filesize > 4 PB ? ;-)
+	getFileSize: function(props) {
+	    var low = props.sizeLow();
+	    if (low < 0) low = low + Math.pow(2, 32);
```
kevodwyer: could extract the result of Math.pow()

PR #73, opened 2019-03-05
Implement bip39 password generator

ianopolous:
This is 8 memorable words for 88 bits of entropy in password
The field hides the password unless generate is clicked


ianopolous:
screenshot_2019-03-05 peergos

PR #74, opened 2019-03-10
streaming now handles partial requests and seeking

kevodwyer:
Chrome 72 broke streaming as you could no longer just push bytes through to the response object (chrome wants to read metadata first - even if it is found at the end of the file)


ianopolous:
Did you measure a speed up with the 1mb initial load? I ask because it will still need to download 5mb and decrypt it to only get 1mb.

PR #76, opened 2019-03-27
[Issue 381] Feedback mechanism

cesarosum:
Created a button and feedback form. 

Please test with corresponding branch in Peergos repo.


src/peergos.css
```diff
@@ -97,6 +97,29 @@ table {
                                   not supported by any browser */
 }
 
+@import url(https://fonts.googleapis.com/css?family=Roboto:400,500,700,300,100);
+
+.feedback-button{
```
ianopolous: can remove this after discussion

src/peergos.css
```diff
@@ -97,6 +97,27 @@ table {
                                   not supported by any browser */
 }
 
+.feedback-button{
+    height:40px;
```
ianopolous: it looks like all this new css isn't used either so presumably it can be removed?

cesarosum: It's used on line 108 of `filesystem.html`.

PR #77, opened 2019-03-29
misc fixes and cleanup

kevodwyer:


vendor/priors/gwt.js
```diff
@@ -6,7 +6,9 @@ function convertToByteArray(target) {
     target.typeMarker = source.typeMarker;
     target.__elementTypeCategory$ = source.__elementTypeCategory$;
     target.__elementTypeId$ = source.__elementTypeId$;
+    var len = target.length;
     target.__proto__ = source.__proto__;
+    target.length = len;
```
ianopolous: this appears to do nothing?

kevodwyer: The line 'target.__proto__ = source.__proto__;' sometimes smashes the length parameter (sets it back to zero). That is why i save it before the destructive behavior. You can see it in action if you put a breakpoint in the function and then download a file. In other scenarios where this function is called the length is preserved. I have no good explanation on the why.

ianopolous: Okay can you put that in a comment as it's highly non obvious from the code.

PR #80, opened 2019-04-09
rework sharedWith cache

kevodwyer:


ianopolous:
can you update the peergos lib otherwise this will break

kevodwyer:
done

PR #82, opened 2019-04-11
check mime type when viewing media

kevodwyer:


kevodwyer:
change much more complicated then required

PR #83, opened 2019-04-11
video stream check

kevodwyer:


src/components/gallery/index.js
```diff
@@ -82,7 +82,7 @@ module.exports = {
             var props = file.getFileProperties();
             var that = this;
             this.showSpinner = true;
-            if(that.supportsStreaming()) {
+            if(that.supportsStreaming() && that.isVideo(file)) {
```
ianopolous: what about audio? We should stream them as well as e.g. podcasts can get very large. Or maybe only for audio > 5MiB?

kevodwyer: fixed

PR #87, opened 2019-04-26
focus input during mkdir

ianopolous:


ianopolous:
I just copied it from login :-)

PR #108, opened 2019-06-25
first pass at admin handling of space requests

ianopolous:


ianopolous:
Time is of the essence

PR #117, opened 2019-07-18
implement history using url fragments to fix the back button issue

ianopolous:
This fixes the back button problem. Now every dir change is pushed encrypted to the url hash, and we listen for hash changes and update the path accordingly.

Solves https://github.com/Peergos/Peergos/issues/503

### DIFF
--- a/vendor/priors/gwt.js
+++ b/vendor/priors/gwt.js
@@ -13,11 +13,11 @@ function convertToByteArray(target) {
 }
 
 function propsToFragment(props) {
-    return "#" + JSON.stringify(props);
+    return JSON.stringify(props);
 }
 
 function fragmentToProps(fragment) {
-    var decoded = decodeURIComponent(fragment.substring(1));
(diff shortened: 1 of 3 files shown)


vendor/priors/gwt.js
```diff
@@ -12,6 +12,15 @@ function convertToByteArray(target) {
     return target;
 }
 
+function propsToFragment(props) {
```
kevodwyer: I'm not sure i like the word fragment also both these functions are just 1 liners. Does it make sense to even have functions rather then call inline?

ianopolous: Fragment is standard and I'm trying to avoid confusion with cryptographic hashes: https://en.wikipedia.org/wiki/Fragment_identifier

These functions are called from multiple components, and I don't want to risk making one out of sync with the other.

kevodwyer: That was new to me! all ok.

PR #134, opened 2019-12-25
Fix/typedarrays

ianopolous:


vendor/priors/gwt.js
```diff
@@ -82,13 +82,13 @@ function postProm(url, data) {
...
 	};
-	
-	req.send(new Uint8Array(data));
+
+	req.send(data);
     }).then(function(result, err) {
         if (err != null)
             future.completeExceptionally(java.lang.Throwable.of(err));
         else
-            future.complete(peergos.shared.user.JavaScriptPoster.convertToBytes(result));
```
kevodwyer: Can we get rid of JavscriptPoster.convertToBytes as we should not be using it anymore?

kevodwyer: can also get rid of arrayutils object in gwt.js also.

ianopolous: I think you're looking at the old version. That's not there.

PR #137, opened 2020-01-02
small ui fixes

kevodwyer:
- signup password safety tickbox must be ticked when creating account
- cancel spinner after navigating to root directory

### DIFF
--- a/src/components/filesystem/index.js
+++ b/src/components/filesystem/index.js
@@ -426,6 +426,9 @@ module.exports = {
 
         goHome: function() {
             this.changePath("/");
+            Vue.nextTick(function() {
+                this.showSpinner = false;
+            }.bind(this));
         },
 
         askMkdir: function() {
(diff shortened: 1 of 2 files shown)


src/components/filesystem/index.js
```diff
@@ -426,6 +426,9 @@ module.exports = {
 
         goHome: function() {
             this.changePath("/");
+            Vue.nextTick(function() {
```
ianopolous: Why isn't this being done the same as any other path change?

kevodwyer: On subsequent 'home' button clicks I can see that vue.js is creating a spinner after the directory is re-calculated. Put breakpoint on existing line console.debug('Spinner module created!'). I don't know why. Adding the Vue.nextTick line ensures that the spinner is hidden again.

ianopolous: It doesn't make sense to me why this would only affect /.

kevodwyer: fixed a different way now

ianopolous: Nice!

PR #139, opened 2020-01-14
file upload improvements

kevodwyer:
splits upload into 2 steps
1. create directory tree and check if files already exists
2. do the upload
see corresponding file-upload-improvements peergos branch.

### DIFF
--- a/src/components/filesystem/index.js
+++ b/src/components/filesystem/index.js
@@ -72,6 +72,7 @@ module.exports = {
             showError:false,
 	    hideDonate: false,
             showSpinner: true,
+            spinnerMessage: '',
             onUpdateCompletion: [] // methods to invoke when current dir is next refreshed
         };
     },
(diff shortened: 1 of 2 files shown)


src/components/filesystem/index.js
```diff
@@ -758,81 +815,50 @@ module.exports = {
...
-            var java_reader = new peergos.shared.user.fs.BrowserFileReader(reader);
             var that = this;
-            var context = this.getContext();
 
             var updateProgressBar = function(len){
                 progress.done += len.value_0;
-                that.progressMonitors.sort(function(a, b) {
-                  return Math.floor(b.done / b.max) - Math.floor(a.done / a.max);
-                });
+                //that.progressMonitors.sort(function(a, b) {
```
ianopolous: Why is this commented? I'd like to see the most progressed one on top, as otherwise it can be off screen

ianopolous: I'm merging we can address this in another PR

PR #142, opened 2020-03-18
video capture

kevodwyer:
paired with video-capture branch in peergos repo


kevodwyer:
Currently only available in Safari Tech Preview. Not yet available in iOS.  https://caniuse.com/#search=MediaRecorder

src/components/filesystem/index.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
             normalSortOrder: true,
             clipboard:{},
             selectedFiles:[],
+            captureChunks:[],
```
ianopolous: how big can this get?

src/components/video-capture/video-capture.html
```diff
@@ -0,0 +1,24 @@
+<transition name="modal">
+<div class="modal-mask" @click="close">
+    <div style="height:30%"></div>
```
ianopolous: would this be better as fullscreen? especially on mobile?

src/components/video-capture/index.js
```diff
@@ -0,0 +1,80 @@
+module.exports = {
+    template: require('video-capture.html'),
+    data: function() {
+        return {
+            buttonLabel: "start capture",
+            captureFilename: null,
+            timesliceInMs: 5000,
+            fps: 3,
```
ianopolous: this sounds like a very low framerate?

src/components/filesystem/index.js
kevodwyer: each chunk is 5 seconds worth of data. The array decreases in size after each chunk is written to storage.

src/components/video-capture/video-capture.html
kevodwyer: Yep, it needs tidy up. Something in the CSS means the view is not sized correctly.

src/components/video-capture/index.js
kevodwyer: It seems to be ignored anyway. I think this PR should be left in stasis until MacOS implementation is available.

PR #143, opened 2020-03-18
fix double opening of apps

kevodwyer:
- first by gallery() then onurlchange()


kevodwyer:
see extra if statements in gallery: function()

PR #146, opened 2020-03-31
update show donate msg logic

kevodwyer:


src/components/filesystem/index.js
```diff
@@ -1508,7 +1513,13 @@ module.exports = {
 	},
 	
 	showDonate: function() {
-	    return this.isLoggedIn && this.usageBytes > 1024*1024 && !this.hideDonate;
+        if (! this.isLoggedIn || this.hideDonate) {
+            return false;
+        }
+        if (this.paymentProperties.paymentServerUrl != undefined && this.paymentProperties.isPaid()) {
```
ianopolous: if we need the null check here that's a bug in paymentproperties

kevodwyer: no bug, it is because showDonate is called before the async code in init() sets the paymentProperties.

kevodwyer: a valid question is: do we want to still display the donate button is there is a payment server but this user has not chosen to set up a DD?

ianopolous: I think you've made the right choice for now. Don't show it on a paid server. We can always reconsider it later.

PR #147, opened 2020-04-08
fix handling of upload failure

kevodwyer:


kevodwyer:
there were 2 problems.
1. In JS the object is not actually throwable and so does not have a printStackTrace() method
2. progressBar was not in scope.

PR #151, opened 2020-05-04
code-editor concurrent changes

kevodwyer:
There is a corresponding peergos pr: https://github.com/Peergos/Peergos/pull/673

### DIFF
--- a/src/components/code-editor/index.js
+++ b/src/components/code-editor/index.js
@@ -135,23 +135,22 @@ module.exports = {
 	    var bytes = convertToByteArray(new TextEncoder().encode(text));
 	    var java_reader = peergos.shared.user.fs.AsyncReader.build(bytes);
 	    const context = this.context;
-	    const size = this.currentFile.getFileProperties().sizeLow();
+	    const existingSizeLo = this.currentFile.getFileProperties().sizeLow() ;
+	    const newSizeLo = existingSizeLo + (Math.max(0, text.length - existingSizeLo));
 	    const that = this;
-
-        this.currentFile.overwriteSectionJS(java_reader, 0, 0, (size - (size % Math.pow(2, 32)))/Math.pow(2, 32), size,
+        this.currentFile.overwriteSectionJS(java_reader, 0, 0, 0, newSizeLo,
             context.network, context.crypto, len => {})
         .thenApply(function(updatedFile) {
             that.saving = false;
             that.currentFile = updatedFile;
         }).exceptionally(function(throwable) {
-            if (throwable.detailMessage.includes("CAS exception")) {
+            if (throwable.detailMessage.includes("Couldn%27t+update+mutable+pointer%2C+cas+failed")) {
                 that.showMessage("Concurrent modification detected", "The file: '" +
                 that.file.getName() + "' has been updated by another user. Your changes could not been saved.");
             } else {
                 that.showMessage("Unexpected error", throwable.detailMessage);
                 console.log('Error uploading file: ' + that.file.getName());
                 console.log(throwable.getMessage());
-                throwable.printStackTrace();
             }
             that.saving = false;
         });
--- a/src/components/code-editor/index.js
+++ b/src/components/code-editor/index.js
@@ -5,70 +5,47 @@ module.exports = {
             showSpinner: false,
 	        expectingSave: false,
 	        saving: false,
-	        showReplace: false,
-            replace_message: "",
-            replace_body: "",
-            replace_consumer_cancel_func: () => {},
-            replace_consumer_func: () => {},
             currentFile: null,
             currentParent: null
         }
     },
-    props: ['context', 'file', 'parent'],
+    props: ['context', 'file', 'parent', 'messages'],
     created: function() {
         let that = this;
         that.currentParent = this.parent;
         that.currentFile = this.file;
-        this.reload(function() {
-            that.startListener(false);
-        });
-    },
-    methods: {
-    reloadFile: function() {
-        let that = this;
-        this.reload(function() {
-            that.startListener(true);
-        });
-    },
-    reload: function(callback) {
-        let that = this;
         this.currentParent.getLatest(this.context.network).thenApply(function(updatedParent) {
             that.currentParent = updatedParent;
             updatedParent.getChild(that.currentFile.getName(), that.context.crypto.hasher, that.context.network).thenApply(function(updatedFile) {
                 that.currentFile = updatedFile.get();
-                callback();
+                that.startListener();
             });
         });
     },
-	startListener: function(reload) {
+    methods: {
+	startListener: function() {
 	    var that = this;
 	    var iframe = document.getElementById("editor");
 	    if (iframe == null) {
-    		setTimeout(function(){that.startListener(reload);}, 1000);
+    		setTimeout(function(){that.startListener();}, 1000);
 	    	return;
 	    }
-	    if (reload) {
-	        iframe.src = "apps/code-editor/index.html";
-    		setTimeout(function(){that.startListener(false);}, 1000);
-	    	return;
-        } else {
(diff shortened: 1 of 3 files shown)


src/components/code-editor/index.js
```diff
@@ -4,43 +4,54 @@ module.exports = {
...
         }
     },
-    props: ['context', 'file', 'parent'],
+    props: ['context', 'file', 'parent', 'messages'],
     created: function() {
-        this.startListener();
+        let that = this;
+        that.currentParent = this.parent;
+        that.currentFile = this.file;
+        this.currentParent.getLatest(this.context.network).thenApply(function(updatedParent) {
```
ianopolous, line 17: This shouldn't be done via the parent. That's what I meant about cleaning up the FileWrapper code. We should be able to modify a file from the file itself. This make it easier to reason about things for example when the file is a different writer to its parent

src/components/code-editor/index.js
```diff
@@ -104,53 +115,72 @@ module.exports = {
...
+	    const file = this.currentFile;
 	    const context = this.context;
 	    const size = text.length;
-	    const parent = this.parent;
 	    const that = this;
-	    this.saving = true;
-	    this.parent.uploadFileJS(file.getName(), java_reader, (size - (size % Math.pow(2, 32)))/Math.pow(2, 32), size,
-				true, true, context.network, context.crypto, len => {}, context.getTransactionService())
-        .thenApply(function(updated) {
+        this.currentParent.uploadFileJS(file.getName(), java_reader, (size - (size % Math.pow(2, 32)))/Math.pow(2, 32), size,
```
ianopolous, line 150: this is where we should use a method on the file itself, rather than the parent

ianopolous: This is also not doing a CAS

kevodwyer: When i run a test highlighting a conncurrent modification the upload fails with a CAS exception in setPointer. if i try again it succeeds - no idea why.  i'm going to need your assistance.

ianopolous: We need a new method on fileWrapper - overwriteSection(long offset, Snapshot expectedVersion, data ...) which uses the cas of the existing version chunks in the champ

src/components/code-editor/index.js
```diff
@@ -143,24 +134,22 @@ module.exports = {
...
+
+        this.currentFile.overwriteSectionJS(java_reader, 0, 0, (size - (size % Math.pow(2, 32)))/Math.pow(2, 32), size,
+            context.network, context.crypto, len => {})
+        .thenApply(function(updatedFile) {
+            that.saving = false;
+            that.currentFile = updatedFile;
         }).exceptionally(function(throwable) {
-            if (throwable.detailMessage.includes("java.lang.JsException: Couldn%27t+update+mutable+pointer%2C+cas+failed%3A+")) {
-                that.showMessage("Concurrent modification detected", "The file: '" + file.getName() + "' has been updated by another user. Your changes could not been saved.");
+            if (throwable.detailMessage.includes("CAS exception")) {
```
ianopolous, line 147: we still need the previous cas error which is a different type of error

src/components/code-editor/index.js
```diff
@@ -143,24 +134,22 @@ module.exports = {
 	    this.saving = true;
 	    var bytes = convertToByteArray(new TextEncoder().encode(text));
 	    var java_reader = peergos.shared.user.fs.AsyncReader.build(bytes);
-	    const file = this.currentFile;
 	    const context = this.context;
-	    const size = text.length;
+	    const size = this.currentFile.getFileProperties().sizeLow();
```
ianopolous, line 138: maybe have sizeLow and sizeHi?

src/components/code-editor/index.js
ianopolous: It means something else in the same writing space was changed, so in principle we should be able to handle it automatically though.

ianopolous: Actually yeah, let's leave it out, and assume we'll handle it internally, by retrying.

src/components/code-editor/index.js
```diff
@@ -143,24 +134,22 @@ module.exports = {
...
-        this.currentParent.uploadFileJS(file.getName(), java_reader, (size - (size % Math.pow(2, 32)))/Math.pow(2, 32), size,
-                true, true, context.network, context.crypto, len => {}, context.getTransactionService())
-        .thenApply(function(updatedParent) {
-            updatedParent.getChild(file.getName(), context.crypto.hasher, context.network).thenApply(function(updatedFile) {
-                that.currentFile = updatedFile.get();
-                that.currentParent = updatedParent;
-                that.saving = false;
-            });
+
+        this.currentFile.overwriteSectionJS(java_reader, 0, 0, (size - (size % Math.pow(2, 32)))/Math.pow(2, 32), size,
```
ianopolous, line 141: Will this extend the file if we write beyond the end?

src/components/code-editor/index.js
```diff
@@ -4,43 +4,45 @@ module.exports = {
...
+        this.currentFile = this.file;
         this.startListener();
     },
     methods: {
 	startListener: function() {
+	    var that = this;
 	    var iframe = document.getElementById("editor");
 	    if (iframe == null) {
-    		setTimeout(this.startListener, 500);
+    		setTimeout(function(){that.startListener();}, 1000);
```
ianopolous, line 21: do we need the function wrapper?

src/components/code-editor/index.js
```diff
@@ -104,53 +106,70 @@ module.exports = {
...
 
-	    this.file.getInputStream(this.context.network, this.context.crypto, props.sizeHigh(), props.sizeLow(), function(read){})
+	    this.currentFile.getInputStream(this.context.network, this.context.crypto, props.sizeHigh(), props.sizeLow(), function(read){})
 	        .thenCompose(function(reader) {
                 var size = that.getFileSize(props);
                 var data = convertToByteArray(new Int8Array(size));
                 return reader.readIntoArray(data, 0, data.length)
                     .thenApply(function(read){
-                        iframe.contentWindow.postMessage({modes:modes, mime:mimeType, readOnly:readOnly, text:new TextDecoder().decode(data)}, '*');
+                        setTimeout(function(){
```
ianopolous, line 117: why was the settimeout needed?

src/components/code-editor/index.js
kevodwyer: yes we do. I have been caught out by that before. For fun test it out yourself. Without the function it will call print() immediately
 	setTimeout(print("hello"), 10000);
      function print(msg) {
      	console.log("msg:" + msg);
      }

src/components/code-editor/index.js
kevodwyer: Yes. I was occasionally seeing an empty iframe. This seemed to ensure the iframe got the content.

src/components/code-editor/index.js
ianopolous: I disagree. print() is a function call, print is a function pointer.

src/components/code-editor/index.js
ianopolous: iframe was null?

src/components/code-editor/index.js
kevodwyer: ahh - yes

kevodwyer: reverted back to original

src/components/code-editor/index.js
kevodwyer: no. felt like a race condition

ianopolous: An empty iframe or an empty document? Could this be a fix for https://github.com/Peergos/Peergos/issues/596 ?

kevodwyer: I am hoping it fixes #596

PR #153, opened 2020-05-27
remove async encryption in web worker

kevodwyer:


kevodwyer:
peergos PR: https://github.com/Peergos/Peergos/pull/689

PR #156, opened 2020-06-13
merge sharing screens

kevodwyer:
peergos PR: https://github.com/Peergos/Peergos/pull/693


ianopolous:
Can you post a screenshot what the new modal looks like?

src/components/filesystem/index.js
```diff
@@ -317,21 +316,21 @@ module.exports = {
 		this.showHexViewer = true;
 	},
 
-	updateCurrentDir: function() {
+	updateCurrentDir: function(refreshSharing) {
```
ianopolous, line 319: I'm a little nervous about coupling the sharing update to the dir directly like this. There are also other calls to this method that are not passing an arg now.

kevodwyer:
> Can you post a screenshot what the new modal looks like?

sharescreen

src/components/filesystem/index.js
kevodwyer: It is only relevant in the one case when forceSharedWithUpdate is triggered in order to refresh the sharedWithData data structure.

PR #157, opened 2020-06-16
merge sharing overlays

kevodwyer:
Peergos PR: https://github.com/Peergos/Peergos/pull/694


kevodwyer:
screenshot

src/components/filesystem/filesystem.html
```diff
@@ -73,22 +72,16 @@
   <share 
     v-if="showShare" 
     v-on:hide-share-with="showShare = false"
-    v-on:update-shared="forceSharedWithUpdate++" 
+    v-on:update-shared="forceSharedRefreshWithUpdate++"
+    v-on:update-shared-refresh="forceSharedRefreshWithUpdate++"
```
ianopolous, line 76: if these do the same thing, why do we need both update-shared and update-shared-refresh events?

kevodwyer: I was hoping to use 'update-shared' for the case where read access was granted. But in testing i found subsequent grant/revoke calls could fail (stale references to file, file's parent). It might be possible to work-around the issue so i left the two attributes there so it can be revisited in the future.

ianopolous: I think we can resolve that issue. Can you describe what steps would cause an issue?

kevodwyer: I did some further testing. the CAS error is not actually related to this PR. Scenario is related to granting 'edit' access for the first time.

PR #158, opened 2020-06-18
add tokenfield ui element to social and share screens

kevodwyer:
using a fork of: https://github.com/sliptree/bootstrap-tokenfield

Found at: https://github.com/Open-Xchange-Frontend/bootstrap-tokenfield
The original project seems to be abandoned	
The file in /dist/bootstrap-okenfield.min.js is still however 4 years old.
There have been more recent commits then the original project.


kevodwyer:
Not ready for merge.
Issues:
Styling is wrong on the tokenfield input field on the share + social screen
Username list is never refreshed - Therefore names not available on social page
Followers list is only refreshed after navigating to the social screen - Therefore names not available on share screen.

ianopolous:
It should end up in vendor.js

kevodwyer:
> It should end up in vendor.js

Fixed.

src/components/filesystem/index.js
```diff
@@ -353,7 +352,7 @@ module.exports = {
             });
         },
 
-	updateSocial: function() {
+	updateSocial: function(callbackFunc) {
```
ianopolous, line 355: Do you think this would be cleaner by making this function return a promise, then you can .then in external change?

ianopolous:
The typeahead needs to match on any place in the name
e.g.
hampshire should match for shire

ianopolous:
What is the ordering of matches? Ideally it would be be levenshtein distance from the query.

kevodwyer:
the suggestion matching is now using the same algorithm as the original typeahead code.

src/components/filesystem/index.js
```diff
@@ -1021,11 +1026,22 @@ module.exports = {
...
             let latestFile = this.files.filter(f => f.getName() == filename)[0];
             this.selectedFiles = [latestFile];
-            var allSharedWithUsernames = this.getContext().sharedWith(latestFile);
-            var read_usernames = allSharedWithUsernames.left.toArray([]);
-            var edit_usernames = allSharedWithUsernames.right.toArray([]);
-            this.sharedWithData = {read_shared_with_users:read_usernames, edit_shared_with_users:edit_usernames};
-            this.showShare = true;
+            let that = this;
+            this.showSpinner = true;
+            this.updateSocial(function(result) {
```
ianopolous, line 1031: we need to avoid this and teh getbypath call. There should be no network requests to show share modal

PR #161, opened 2020-07-09
Feat/message bar

kevodwyer:
This turns the existing feedback functionality into a means of providing user <--> server messaging.


kevodwyer:
messages

kevodwyer:
feedback

kevodwyer:
messages

src/components/feedback/feedback.html
```diff
@@ -3,19 +3,67 @@
     <meta http-equiv="content-type" content="text/html; charset=utf-8" />
     <div style="height:30%"></div>
     <div class="modal-container" @click.stop style="height:70%;overflow-y:auto">
+        <span @click="close" class="close">&times;</span>
         <div class="modal-header">
-            <h2>Feedback</h2>
+            <div v-if="isFeedback" >
```
ianopolous, line 8: using feedback component for messages as well feels a little wrong to me. If we're going to do it this way, then we should generify it a bit more. So have the title as a property for example. Essentially try and remove all the isFeedback branches.

src/components/filesystem/index.js
```diff
@@ -196,14 +199,43 @@ module.exports = {
             }
         });
             }
+        this.showPendingMessages();
 	},
-	
-        processPending: function() {
-            for (var i=0; i < this.onUpdateCompletion.length; i++) {
-                this.onUpdateCompletion[i].call();
+	showPendingMessages: function() {
```
ianopolous, line 204: can we call this showPendingServerMessages? I'm trying to keep the naming clear and reserve message on its own for keymail.

src/components/filesystem/index.js
```diff
@@ -887,6 +919,144 @@ module.exports = {
...
+                that.showSpinner = false;
+            });
+        },
+
+        sendMessage: function(msgId, contents) {
+            let that = this;
+            let message = this.getMessage(msgId);
+            if (message != null) {
+                this.showSpinner = true;
+                var trimmedContents = contents.length > 1000 ? contents.substring(0, 1000) : contents;
```
ianopolous, line 973: we define max length in the ServerMessage class. Can we reference that directly? Does it need a jsannotation?

kevodwyer: done

src/components/filesystem/index.js
kevodwyer: done

src/components/feedback/feedback.html
kevodwyer: done

PR #163, opened 2020-07-14
Fix/s3 storage external user

ianopolous:


vendor/priors/gwt.js
```diff
@@ -25,37 +25,28 @@ function fragmentToProps(fragment) {
 function getProm(url) {
```
kevodwyer, line 25: what about the other methods that use Promise?

vendor/priors/gwt.js
```diff
@@ -75,7 +66,11 @@ function postProm(url, data) {
 		resolve(new Int8Array(req.response));
             }
             else {
-		reject(req.getResponseHeader("Trailer"));
+		try {
+		    reject(req.getResponseHeader("Trailer"));
+		} catch (e) {
+		    reject("Error");
```
kevodwyer, line 72: how about adding the actual exception to the error message?

vendor/priors/gwt.js
ianopolous: There's nothing wrong with the promise, it's just unnecessary

vendor/priors/gwt.js
ianopolous: done

PR #164, opened 2020-07-15
update wording on signup in warning

ianopolous:


cesarosum:
I think it's too wordy.  How about the following with a link to an FAQ entry? I'm happy to write it.

I understand that passwords cannot be reset or recovered by Peergos. If I forget my password or it is stolen, I will permanently lose access to my account and data. Click here[link] for more information.

PR #166, opened 2020-07-22
improve watcher to handle when file is updated

kevodwyer:
This is a consequence of peergos PR: https://github.com/Peergos/Peergos/pull/719

### DIFF
--- a/src/components/code-editor/index.js
+++ b/src/components/code-editor/index.js
@@ -5,7 +5,8 @@ module.exports = {
             showSpinner: false,
 	        expectingSave: false,
 	        saving: false,
-	        currentFile: null
+	        currentFile: null,
+	        isFileWritable: null
         }
     },
     props: ['context', 'file', 'messages'],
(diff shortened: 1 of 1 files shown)


src/components/code-editor/index.js
```diff
@@ -168,7 +169,10 @@ module.exports = {
     },
     computed: {
         isWritable: function() {
-	        return this.currentFile.isWritable();
+            if (this.isFileWritable == null) {
```
ianopolous, line 172: Can you explain why this is necessary?

kevodwyer: I was seeing "This file has already been modified, use the returned instance"

kevodwyer: consequence of the call to clean()

ianopolous: Surely then we should be updating currentFile?

kevodwyer: yes - that happens on line 145

ianopolous: then why is this necessary?

kevodwyer: steps
1. computed field is called for the first time, isWriteable() method called on currentFile
2. file contents loaded into editor.
3. user makes change and select save
At this point computed field is called again. Previously this would lead to the error seen above even though currentFile is updated after the call to overwriteFileJS().

ianopolous: Hmm it still feels like an antipattern. I suggest we try and rework it to not call it on the old object during an update.

ianopolous: I would use the same pattern we do in filesystem, watchers for new currentFile values update a writable property.

kevodwyer: I have removed the watcher function. It isn't required.

PR #167, opened 2020-07-29
[fix_browser_tests]

cboddy:
This enables the browser tests to be run on new pushed commits via Github actions.
NB: chrome-84 is now _required_ since this is the first version that can both be
run headlessly and download a file without hitting various bugs in the
chromedriver. Currently, only one test is run that signs up a new user and uploads
and downloads a file and checks the contents. The rest of the tests are skipped as
they need to be updated to reflect recent changes in this repo.

More details below:

* fix signup controller and update to python-3
* bump max-users to 100
* Added profile to selenium driver to disable the  file-system app that pops up when downloading a file
* Updated file-upload
* Fixed file-download
* Added smoke test that uploads and downloads a file
* skip broken tests - these need to be updated to reflect the UI changes (mainly changed element ids) since  these were last updated
* update browser-test scripts and  readme
* add github ci for browser tests
* add update docs


ianopolous:
Awesome! 

A few suggestions:
1. Could we turn off the server log (it's not useful debugging ui issues in practice)
2. Could we retrieve the browser console and print it when there is a failure

build.xml
```diff
@@ -201,7 +201,7 @@ Class-Path: .</echo>
       <arg value="-webcache"/>
       <arg value="false"/>
       <arg value="-max-users"/>
-      <arg value="5"/>
+      <arg value="100"/>
```
ianopolous, line 204: This is set low so it's easy to test the behaviour when the server is full. (Which eventually we should have browser tests for). Seems like the browser tests should invoke the jar directly with this higher value and no log.

cboddy: I think we do need a different way to start peergos since ant  also has the spurious "BUILD FAILED" message.

cboddy:
> Awesome!
> 
> A few suggestions:
> 
>     1. Could we turn off the server log (it's not useful debugging ui issues in practice)
> 
>     2. Could we retrieve the browser console and print it when there is a failure
> 
>     3. Could we download the chrome driver rather than check it in, as it will be changing frequently to match chromium version

1) Yes - will do
2) Good idea - should be possible. Let's push that to a different PR
3) I think the frequency at which we'll update the chromedriver  (and chrome on all of our local  lachines) should be  pretty low, like every 6 months which I'm personally  OK with.  If it becomes an issue we can do as you suggest later

ianopolous:
We've had 3 major chrome versions released just this month. And we do stay bleeding (stable) edge because we need new features and fixes.

PR #168, opened 2020-07-30
Update UI to re-sync with sharing changes

kevodwyer:


src/components/filesystem/index.js
```diff
@@ -331,25 +332,43 @@ module.exports = {
                 that.currentDir = file.get();
                 that.updateFiles(refreshSharedWith);
             }).exceptionally(function(throwable) {
-                throwable.printStackTrace();
+                console.log(throwable.getMessage());
```
ianopolous, line 335: any reason to not also do a stack trace?

kevodwyer: stacktrace not available in js

ianopolous: I've seen java stacktraces in the browser console?

src/components/filesystem/index.js
```diff
@@ -331,25 +332,32 @@ module.exports = {
...
-            current.getChildren(that.getContext().crypto.hasher, that.getContext().network).thenApply(function(children){
-                var arr = children.toArray();
-                that.showSpinner = false;
-                that.files = arr.filter(function(f){
-                    return !f.getFileProperties().isHidden;
+            let that = this;
+            let context = this.getContext();
+            let directoryPath = peergos.client.PathUtils.directoryToPath(that.path);
+            context.getDirectorySharingState(directoryPath).thenApply(function(updatedSharedWithState) {
+                that.sharedWithState = updatedSharedWithState;
```
ianopolous, line 346: I would wait to set this until the children are also retrieved

src/components/filesystem/index.js
```diff
@@ -331,25 +332,32 @@ module.exports = {
...
+                current.getChildren(that.getContext().crypto.hasher, context.network).thenApply(function(children){
+                    var arr = children.toArray();
+                    that.showSpinner = false;
+                    that.files = arr.filter(function(f){
+                        return !f.getFileProperties().isHidden;
+                    });
+                    if (refreshSharedWith) {
```
ianopolous, line 353: when do we not want to refreshSharedWith?

src/components/filesystem/index.js
kevodwyer: There is no printStackTrace() on the throwable object returned at least. Seems it isn't exposed. We don't call printStackTrace anywhere in the vanilla js code.

src/components/filesystem/index.js
kevodwyer: Simplified now

ianopolous: Very nice

PR #170, opened 2020-08-05
allow server messages to safely word wrap

ianopolous:
Screenshot_2020-08-05 Peergos

### DIFF
--- a/src/components/feedback/feedback.html
+++ b/src/components/feedback/feedback.html
@@ -40,8 +40,11 @@
                         </tr>
                         <tr v-if="message.visible">
                             <td>
-                                <div v-if="message.from == 'FromUser'">
-                                    <pre style="background-color: #ffffff;">{{message.contents}}</pre>
+                                <div v-if="message.from == 'FromUser'" style="background-color: #ffffff;">
+                                  <div v-for="paragraph in message.paragraphs" >
+                                    {{paragraph}}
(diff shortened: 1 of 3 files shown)


src/components/feedback/feedback.html
```diff
@@ -44,7 +44,10 @@ <h3>You can tell us here how we can improve, or you can chat with us on <a href=
                                     <pre style="background-color: #ffffff;">{{message.contents}}</pre>
```
kevodwyer, line 44: What about here?

ianopolous: Fixed

src/components/feedback/feedback.html
```diff
@@ -40,8 +40,11 @@ <h3>You can tell us here how we can improve, or you can chat with us on <a href=
                         </tr>
                         <tr v-if="message.visible">
                             <td>
-                                <div v-if="message.from == 'FromUser'">
-                                    <pre style="background-color: #ffffff;">{{message.contents}}</pre>
+                                <div v-if="message.from == 'FromUser'" style="background-color: #ffffff;">
```
kevodwyer, line 43: A simplification would be to move the if statement into the style (I wish i did that when i originally wrote it).

ianopolous: what is the syntax for that?

PR #176, opened 2020-08-30
upload new/empty file menu item from context menu

kevodwyer:
This is twined with peergos PR: https://github.com/Peergos/Peergos/pull/734

### DIFF
--- a/src/components/filesystem/index.js
+++ b/src/components/filesystem/index.js
@@ -1514,7 +1514,43 @@ module.exports = {
             }.bind(this));
             e.preventDefault();
         },
-
+        createTextFile: function() {
+            this.closeMenu();
+            this.prompt_placeholder='File name';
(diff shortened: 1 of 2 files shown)


src/components/filesystem/index.js
```diff
@@ -1514,7 +1514,43 @@ module.exports = {
             }.bind(this));
             e.preventDefault();
         },
-
+        createTextFile: function() {
+            this.closeMenu();
+            this.prompt_placeholder='File name';
+            this.prompt_message='Enter a file name';
+            this.prompt_value='';
```
ianopolous, line 1521: maybe default to example.txt?

kevodwyer: But then the prompt_message would not be displayed.
prompt

ianopolous (approved):

PR #186, opened 2020-11-01
move icons around

ianopolous:


kevodwyer (commented):
selecting an item on app menu should make menu disappear by calling toggleNav(). That is the behaviour for the todo board.

kevodwyer (approved):

PR #192, opened 2020-11-16
support multiple calendars

kevodwyer:
a user can have multiple calendars
each calendar is totally separate
each calendar is associated to a name and a color
both the name and color is customisable
a calendar can both be added and deleted
allow import of an ical file directly into a calendar

sharing a calendar is not in scope for this PR

### DIFF
--- a/assets/apps/calendar/init.js
+++ b/assets/apps/calendar/init.js
@@ -1129,13 +1129,21 @@ function calendarModalHandler(event, colorChange, configurationPopupCloseFunc) {
     }
     if (event.target.id == "calendarModal") {
         configurationPopupCloseFunc();
-    } else if (event.target.id == "color-picker-container"
-        || event.target.id == "add-calendar-container"
-        || event.target.className == "calendar-modal-content") {
-            calendarColorChange(colorChange);
+    } else if (event.target.id == "color-picker-cancel-btn") {
+        resetColorChange(colorChange);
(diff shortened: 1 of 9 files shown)


assets/apps/calendar/init.js
```diff
@@ -127,32 +134,83 @@ window.addEventListener('message', function (e) {
...
 });
-function renameCalendar(currentName) {
-    mainWindow.postMessage({action:'requestRenameCalendar', currentName: currentName}, origin);
+function renameCalendar(calendar) {
+    mainWindow.postMessage({action:'requestRenameCalendar', calendar: calendar}, origin);
+}
+function addCalendar(){
+    let idAsInt = new Number(CALENDAR_ID_MY_CALENDAR) + CalendarList.length;
+    let color = colorPalette[(idAsInt % 8) + 8];
+    mainWindow.postMessage({action:'requestAddCalendar', newColor:color}, origin);
```
ianopolous, line 160: why do these need to go back to the trusted code? can't we implement them in the app directly?

kevodwyer: I wanted to use the existing dialogs for add, rename. sandboxed apps would use there own.

ianopolous: So long as we're aware we'll need to pull them out once we have a sandbox.

PR #198, opened 2020-11-25
make sure memo change is acted upon

kevodwyer:


assets/apps/calendar/init.js
```diff
@@ -72,9 +72,10 @@ function buildUI(isCalendarReadonly) {
         'beforeUpdateSchedule': function(e) {
             var schedule = e.schedule;
             var changes = e.changes;
-            if(changes == null) {
-                return;
-            }
+            //do not simply return if no changes. Memo text is not recorded as a change!
+            //if(changes == null) {
```
ianopolous, line 76: what was this here for?

kevodwyer: the calendar component stores the changes made in the UI. The Memo field is a custom field we add. It is not part of the calculated changes. It was tempting to short-circuit processing if no change was found. I am making it clear in the code that should not be done.

ianopolous (approved):

assets/apps/calendar/init.js
ianopolous: What the consequence for the user of the short circuit?

kevodwyer: The regression meant the calendar event was not updated when the memo (detailed description) field was updated.

PR #202, opened 2020-11-28
Fix a bunch of layout issues, mostly on mobile

ianopolous:
Also remove the donation button.


kevodwyer (commented):
How about adding a tour page for profile, or even navigating there after sign-up?

ianopolous:
Yep, we will add tour pages for todo, calendar, profile, website hosting and social feed

kevodwyer (approved):

PR #205, opened 2020-12-01
check available space during upload process

kevodwyer:
This is not ideal in the case of multiple uploads
Also handles the case of file copying and directory copying

### DIFF
--- a/src/components/filesystem/filesystem.html
+++ b/src/components/filesystem/filesystem.html
@@ -266,7 +266,7 @@
   <div v-if="progressMonitors.length>0" class="progressholder">
     <progressbar 
         v-for="progress in progressMonitors" v-if="progress.show"
-      v-on:hide-progress="progressMonitors.pop(progress)"
+      v-on:hide-progress="progressMonitors.splice(progressMonitors.indexOf(progress), 1)"
       :title="progress.title" 
       :done="progress.done"
       :max="progress.max" />
--- a/src/components/filesystem/index.js
+++ b/src/components/filesystem/index.js
@@ -1417,7 +1417,8 @@ module.exports = {
                         let newPath = path + "/" + childProps.name;
                         that.calculateDirectorySize(child, newPath, accumulator, future);
                     } else {
-                        accumulator.size += that.getFileSize(childProps);
+                        let size = that.getFileSize(childProps);
+                        accumulator.size += (size + (4096 - (size % 4096)));
                     }
                 }
                 accumulator.walkCounter--;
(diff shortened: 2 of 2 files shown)


src/components/filesystem/index.js
```diff
@@ -270,15 +271,20 @@ module.exports = {
             if (this.isSecretLink)
 		return;
 	    var that = this;
-	    this.context.getSpaceUsage().thenApply(u => that.usageBytes = u.value_0);
+	    this.context.getSpaceUsage().thenApply(u => {
+	        that.usageBytes = u
```
ianopolous, line 275: do we not need the value_0?

ianopolous:
Have you tested it with >4GiB size files, and/or space left?

src/components/filesystem/index.js
kevodwyer: No. We call .toString() on the Long object.

ianopolous: Wasn't there a reason for it? Maybe to do with handling the 4GiB barrier?

kevodwyer: It wasn't consistent in any case (quotaBytes versus usageBytes) . It is now. I have tested it with large numbers.

ianopolous: What sizes have you tested it with?

kevodwyer: Usage: 31.68 KiB / 200 GiB
Usage: 31.67 KiB / 20000 GiB
Usage: 31.67 KiB / 2000000 GiB

kevodwyer:
> Have you tested it with >4GiB size files, and/or space left?

I used a VM image and the numbers seen right

ianopolous:
We need to test where the quota, the file size, and the (quota - file size) are just below and above 4 GiB.

ianopolous:
Fixes: https://github.com/Peergos/Peergos/issues/748

src/components/filesystem/filesystem.html
```diff
@@ -266,7 +266,7 @@
   <div v-if="progressMonitors.length>0" class="progressholder">
     <progressbar 
         v-for="progress in progressMonitors" v-if="progress.show"
-      v-on:hide-progress="progressMonitors.pop(progress)"
+      v-on:hide-progress="progressMonitors.splice(progressMonitors.indexOf(progress), 1)"
```
ianopolous, line 269: great catch! Does this fix the random closing of upload progress when concurrently uploading many files?

kevodwyer: yes. Note also that uploads are no longer concurrent. This makes it possible to check available space after each upload and also should fix problems whereby the ui would get out of sync

kevodwyer:
> Have you tested it with >4GiB size files, and/or space left?

Yes. Works as expected

ianopolous (approved):

PR #206, opened 2020-12-03
enable/disable toolbar buttons when loading events

kevodwyer:


kevodwyer:
also fixes issue with size of calendar and with sharing item name

ianopolous (approved):

PR #211, opened 2020-12-07
workaround for Firefox iframe behaviour

kevodwyer:
Firefox seems to have problems when a page with an iframe is reloaded.
The message sent from page to iframe is lost.


ianopolous:
Does this fix the need to close calendar after importing an event?

kevodwyer:
> Does this fix the need to close calendar after importing an event?

Yes

ianopolous (approved):

PR #225, opened 2021-01-26
whole calendar sharing

kevodwyer:
See Peergos change: https://github.com/Peergos/Peergos/pull/787

Adds the ability to share an entire calendar, either in-band or via secret link
The shared calendar is read-only
The receiver imports the calendar via the timeline


src/components/timeline/index.js
```diff
@@ -10,8 +10,8 @@ module.exports = {
             noMoreResults: false
         }
     },
-    props: ['context','navigateToAction','viewAction', 'messages', 'getFileIcon', 'socialFeed',
-        'importCalendarFile', 'displayProfile'],
+    props: ['context','navigateToAction','viewAction', 'messages', 'getFileIconFromFileAndType', 'socialFeed',
```
ianopolous, line 13: does getFileIconFromFileAndType still return the class? I wouldn't encode the arguments in the method name.

kevodwyer: Yes. 
The problem was the existing getFileIcon would recognise the file as a directory and return the folder class. I wanted the calendar class. So i restructured the function to match on type first (which i pass in as a string) and then file/folder if no match.
In order to not break existing code the function getFileIcon calls through to  getFileIconFromFileAndType using the type from the file.getFileProperties().getType() so no change there.

ianopolous (approved):

PR #226, opened 2021-02-01
group sharing

kevodwyer:
currently limited to friends and followers
a user can have both read & read/write access (either directly or via groups)
if a user is given access at the same level as an individual and as a group, then the individual access is not shown on UI
if a user is given access at the same level as an individual and as a group and the group access is subsequently removed - then their individual access is also removed.
dependant on Peergos PR: https://github.com/Peergos/Peergos/pull/783

### DIFF
--- a/src/components/share/index.js
+++ b/src/components/share/index.js
@@ -18,7 +18,7 @@ module.exports = {
             modalLinks:[]
         }
     },
-    props: ['data', 'friendnames', 'followernames', 'groups', 'files', 'parent', 'path', 'context', 'messages', 'fromApp', 'displayName', 'allowReadWriteSharing', 'allowCreateSecretLink'],
+    props: ['data', 'friendnames', 'followernames', 'groups', 'files', 'path', 'context', 'messages', 'fromApp', 'displayName', 'allowReadWriteSharing', 'allowCreateSecretLink'],
     created: function() {
         Vue.nextTick(this.setTypeAhead);
     },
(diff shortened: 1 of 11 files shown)


src/components/timeline/index.js
```diff
@@ -64,7 +64,8 @@ module.exports = {
             let filteredSharedItems = [];
             for(var i=0; i < items.length; i++) {
                 let currentSharedItem = items[i];
-                if (!currentSharedItem.path.startsWith("/" + currentSharedItem.owner + "/.profile/")) {
+                if (!currentSharedItem.path.startsWith("/" + currentSharedItem.owner + "/.profile/")
+                    && !currentSharedItem.path.startsWith("/" + currentSharedItem.owner + "/shared/.")) { //groups
```
ianopolous, line 68: do you think this would be better to fix by making the dirs hidden?

src/components/share/index.js
```diff
@@ -219,11 +341,17 @@ module.exports = {
         }
 	},
 	    setTypeAhead: function() {
-            let followernames = this.followernames;
+            let allNames = this.followernames.slice();
+            for(var i = 0; i < this.friendnames.length;i ++) {
+                let friend = this.friendnames[i];
+                if (allNames.indexOf(friend) == -1) {
```
ianopolous, line 347: aren't all the friends in the followers?

src/components/share/index.js
```diff
@@ -62,7 +64,30 @@ module.exports = {
...
         },
-
+        gatherAllUsersToUnshare: function(currentSharedWithUsernames, usernamesToUnshare) {
+            let friendGroupUid = this.getGroupUid(peergos.shared.user.SocialState.FRIENDS_GROUP_NAME);
+            let includesFriends = usernamesToUnshare.indexOf(friendGroupUid) > -1;
+            let followersGroupUid = this.getGroupUid(peergos.shared.user.SocialState.FOLLOWERS_GROUP_NAME);
+            let includesFollowers = usernamesToUnshare.indexOf(followersGroupUid) > -1;
+            let allUsers = usernamesToUnshare.concat([]);
+            if (includesFriends) {
+                for(var i = 0; i < currentSharedWithUsernames.length; i++) {
```
ianopolous, line 74: i seems to be indexing two different things here

src/components/timeline/index.js
kevodwyer: I think the setup is fine. The timeline will not display hidden files.  See createTimelineEntry().  The use of this filter method is to short-circuit the gathering of results (expensive operation) so it doesn't even try to createEntries for these items.

src/components/share/index.js
kevodwyer: No. 
They are kept separate when derived in filsystem.js
followerNames = followerNames.filter(x => !friendNames.includes(x));

src/components/share/index.js
kevodwyer: Fixed

PR #235, opened 2021-03-23
fix streaming sha256 remaining length calculation

ianopolous:


vendor/priors/gwt.js
```diff
@@ -383,7 +383,7 @@ var scryptJS = {
 		    }
 		    return asyncReader.readIntoArray(buf, 0, Math.min(buf.length, toRead)).thenCompose(function(bytesRead){
 			h.update(bytesRead == buf.length ? buf : buf.slice(0, bytesRead));
-			return priv.recurse(toRead - bytesRead);
+			return priv.recurse(toRead - bytesRead.value_0);
```
kevodwyer, line 386: I can see other references to bytesRead in the method. They should also use .value_0

ianopolous: Somehow it works?

ianopolous: I've changed it anyway, but will need more testing now.

kevodwyer (approved):

PR #244, opened 2021-03-29
social feed improvements

kevodwyer:
- shared items are now enclosed in the same box as the comments relating to that item
- comments on shared items are only shared with the sharer.


ianopolous (approved):
Nice!

ianopolous (approved):

PR #252, opened 2021-04-28
remove serial only processing in reduce

kevodwyer:


ianopolous (approved):
Fantastic. One small improvement would be to make the recurring event loading parallel to the normal event loading.

PR #302, opened 2021-09-04
Add an identity proof viewer

ianopolous:
This also meant fixing public links to files, and making sure they propagate open=true to the resulting secretlink


src/components/identity-proof-viewer.vue
```diff
@@ -0,0 +1,69 @@
...
+                return;
+            if (this.file.isDirectory())
+                return;
+            var props = this.file.getFileProperties();
+            var that = this;
+            this.showSpinner = true;
+            this.file.getInputStream(this.context.network, this.context.crypto, 
+                props.sizeHigh(), props.sizeLow(), 
+                function(read) {})
+                .thenCompose(function(startReader) {
```
kevodwyer, line 51: This seems overly complicated. Shouldn't it just read the entire file?

src/components/filesystem/index.js
```diff
@@ -514,6 +523,8 @@ module.exports = {
 		this.showCodeEditor = true;
 	    else if (app == "hex")
 		this.showHexViewer = true;
+	    else if (app == "identity-proof")
+		this.showIdentityProof = true;
```
kevodwyer, line 527: what file icon is used?

kevodwyer (commented):
could this be added to edit/view profile page?

src/components/filesystem/index.js
ianopolous: the default is fine for now

src/components/identity-proof-viewer.vue
ianopolous: I've simplified it now

kevodwyer (approved):

PR #315, opened 2021-09-30
Implement minimal PWA requirements

ianopolous:


ianopolous:
We need to understand updates before merging.

ianopolous:
Can't get updates to work reliably. So will wait until PWAs are more mature and more supported.

PR #326, opened 2021-11-08
change password dialog fixes

kevodwyer:
update logic to show an error if the current password is incorrect or the same as the new password


ianopolous (approved):
Nice I also noticed this yesterday. Great!

PR #327, opened 2021-11-08
fix logout on user settings menu

kevodwyer:
call to this.$store.commit("SET_CONTEXT", null) was failing as shallow() was expecting a non-null parameter value


ianopolous (approved):
Great catch!

PR #328, opened 2021-11-12
Check that file operation is possible within account space limit

kevodwyer:


ianopolous:
It looks like both errors say, "File copy operation exceeds Account quota" Should one be talking about upload?

Also I don't think we should use the term quota in the UI (I don't think we currently use it anywhere?). Maybe something like
"File copy operation exceeds total space. Please upgrade to get more space."

kevodwyer:
> It looks like both errors say, "File copy operation exceeds Account quota" Should one be talking about upload?
> 
> Also I don't think we should use the term quota in the UI (I don't think we currently use it anywhere?). Maybe something like "File copy operation exceeds total space. Please upgrade to get more space."

done

src/views/Drive.vue
```diff
@@ -1038,38 +1038,43 @@ module.exports = {
...
-                        that.$toast({component: ProgressBar,props:  progress} , { icon: false , timeout:false, id: files[i].name})
-                        progressBars.push(progress);
-                    }
-                    const uploadParams = {
-                        applyReplaceToAll: false,
-                        replaceFile: false,
-                        filesUploaded: 0
+                    let spaceAfterOperation = this.checkAvailableSpace(totalSize);
+                    if (spaceAfterOperation < 0) {
+                        let errMsg = "File copy operation exceeds available space\n" + "Please free up " + helpers.convertBytesToHumanReadable('' + -spaceAfterOperation) + " and try again";
```
ianopolous, line 1047: is this a file upload or copy?

kevodwyer: upload. have now fixed message

ianopolous (approved):

PR #331, opened 2021-11-26
clear form autocomplete component when value is processed

kevodwyer:
This clears the input field after
- friend request sent 
(also now allow sending friend request if name is not yet known - functionality previously available in old ui)
- member/admin added to chat


ianopolous (approved):
nice

PR #334, opened 2021-11-29
Fix/newsfeed bugs

kevodwyer:
1. sorting of items was potentially wrong
2. replying to an edited social post was not handled correctly
3. walking back in history caused duplicate entries to be shown. (paging will be revised as part of a bigger re-write)


ianopolous (approved):
Excellent

PR #340, opened 2021-12-14
can now open a file by double clicking

kevodwyer:


ianopolous:
Is there a reason not to use the double click event? https://www.w3schools.com/jsref/event_ondblclick.asp

kevodwyer:
> Is there a reason not to use the double click event? https://www.w3schools.com/jsref/event_ondblclick.asp

If i have both a single click handler and double click event handler i will get duplicated events!

ianopolous (approved):

PR #341, opened 2021-12-15
improve thumbnail creation

kevodwyer:
- handle thumbnail creation failure gracefully for images and large videos
- add option to create thumbnail from file menu
- This change relies on a Peergos PR: https://github.com/Peergos/Peergos/pull/882

### DIFF
--- a/src/views/Drive.vue
+++ b/src/views/Drive.vue
@@ -669,7 +669,7 @@ module.exports = {
                 return false;
             let file = this.selectedFiles[0];
             let that = this;
-            file.calculateThumbnail(this.context.network, this.context.crypto).thenApply(res => {
+            file.calculateAndUpdateThumbnail(this.context.network, this.context.crypto).thenApply(res => {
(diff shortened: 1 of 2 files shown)


src/views/Drive.vue
```diff
@@ -384,6 +385,20 @@ module.exports = {
 				return false;
 			}
 		},
+		canCreateThumbnail() {
+            if (this.selectedFiles.length != 1)
+                return false;
+            let file = this.selectedFiles[0];
+            if (file.props.thumbnail.ref != null) {
+                return false;
```
ianopolous, line 393: I can see cases where I might want to recalculate the thumbnail, like when a movie picked a bad frame, or if the thumbnail was made in firefox and I want webp thumbnail (and I'm in chrome)

ianopolous: But maybe those aren't important?

ianopolous (approved):

PR #344, opened 2022-01-02
support block level access control (BATs)

ianopolous:
We just need to ensure we have a mirror bat id at login and pass it to mkdir and uploadFileJS calls

Also implement hmacSha256 on JSScrypt


kevodwyer (commented):
the indentation is a bit off

kevodwyer (approved):

PR #345, opened 2022-01-18
cas exception error message changed at some point

kevodwyer:


src/components/code-editor/index.js
```diff
@@ -172,8 +172,8 @@ module.exports = {
             that.currentFile = updatedFile;
             that.$emit("update-refresh");
         }).exceptionally(function(throwable) {
-            if (throwable.detailMessage.includes("Couldn%27t+update+mutable+pointer%2C+cas+failed")
-                ||   throwable.detailMessage.includes("CAS exception updating cryptree node.")) {
+            if (throwable.detailMessage.includes("CAS exception updating cryptree node.")
+                ||   throwable.detailMessage.includes("Corenode Crypto CAS failed!")) {
```
ianopolous, line 176: this has nothing to do with the corenode. I'll update this Java side. Really we should use a new Exception Type.

kevodwyer: Js code now uses the revised exception message.

ianopolous (approved):

PR #346, opened 2022-01-25
display drag and drop target only after workspace is loaded

kevodwyer:
also only update current dir if old path is set

### DIFF
--- a/src/views/Drive.vue
+++ b/src/views/Drive.vue
@@ -54,8 +54,7 @@
 						:file="file"
 						:itemIndex="index"
 					/>
-					<DriveGridDrop v-if="sortedFiles.length==0 && currentDir != null && currentDir.isWritable()">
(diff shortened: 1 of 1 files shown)


src/views/Drive.vue
```diff
@@ -54,8 +54,7 @@
 						:file="file"
 						:itemIndex="index"
 					/>
-					<DriveGridDrop v-if="sortedFiles.length==0 && currentDir != null && currentDir.isWritable()">
-						Upload a file by dragging and dropping here or clicking the icon
```
ianopolous, line 58: Have we lost this text?

kevodwyer: 'Drop files here' is displayed as currently. Text comes from DriveGridDrop.vue

ianopolous (approved):

PR #350, opened 2022-02-10
fix safari video streaming

kevodwyer:


ianopolous (approved):
Stunning! Brilliant fix!

PR #352, opened 2022-02-19
use new bulk upload @JSMethod

kevodwyer:


src/views/Drive.vue
```diff
@@ -1109,12 +1112,11 @@ module.exports = {
                         //resetting .value tricks browser into allowing subsequent upload of same file(s)
                         document.getElementById('uploadFileInput').value = "";
                         document.getElementById('uploadDirectoriesInput').value = "";
-                        let future = peergos.shared.util.Futures.incomplete();
                         let progressBars = [];
                         for(var i=0; i < files.length; i++) {
                             let that = this;
-                            var thumbnailAllocation = Math.min(100000, files[i].size / 10);
-                            var resultingSize = files[i].size + thumbnailAllocation;
+                            //var thumbnailAllocation = Math.min(100000, files[i].size / 10);
```
ianopolous, line 1118: is the thumbnailAllocation still meant to be commented?

src/views/Drive.vue
```diff
@@ -1211,57 +1265,51 @@ module.exports = {
...
+                            component: ProgressBar,
+                            props:  {
+                            title: progress.title,
+                            done: progress.done,
+                            max: progress.max
+                            },
+                        }
                    });
-
+                //                const thumbnailAllocation = Math.min(100000, file.size / 10);
```
ianopolous, line 1283: same question here

src/views/Drive.vue
kevodwyer: The thumbnailAllocation code was a bit of a hack. Previously we would fill the progress bar to 100% when the upload had finished (post thumbnail creation). No longer get a notification when individual upload finished. I have now removed the code.

src/views/Drive.vue
kevodwyer: code now removed

ianopolous (approved):
Excellent!

PR #354, opened 2022-03-12
fix spinner positioning

kevodwyer:
make sure spinner appears in centre of screen when on profile and change pwd modals

### DIFF
--- a/src/peergos.css
+++ b/src/peergos.css
@@ -571,7 +571,6 @@ h3 {word-wrap:break-word; white-space:normal;}
 }
 
 .flex-item > div {
-  width: 300px;
   padding: 10px;
   text-align: left;
   font-size: 20px;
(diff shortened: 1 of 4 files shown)


src/peergos.css
```diff
@@ -571,7 +571,6 @@ h3 {word-wrap:break-word; white-space:normal;}
 }
 
 .flex-item > div {
-  width: 300px;
```
ianopolous, line 574: does this affect anything else?

kevodwyer: flex-item only used on profile + search page. 
On the profile page the change means the bio & status field is sized more appropriately.  
Have confirmed style is fine on search page.
I also made a small style change on search page to make the search type dropdown more prominent.

ianopolous (approved):

PR #356, opened 2022-03-16
remove duplicate update social call

kevodwyer:


src/views/Social.vue
```diff
@@ -171,17 +171,12 @@ module.exports = {
         }
     },
 	created() {
-		this.updateSocial(this.test);
+		this.updateSocial();
```
ianopolous, line 174: Does this mean it won't be updated when we click on it? I thought that's what mounted was for?

ianopolous: I've specifically tested that before. It certainly should have.

ianopolous: I've just confirmed on master that this still works.

kevodwyer: We must be talking about different things. On current deployment, If i put breakpoints in created/mounted and click the social icon, then i hit the breakpoints. if i click the social icon again while still on the social page then i do not hit the breakpoints and the page is not reloaded.

ianopolous: I'm talking about updating the social data each time we open the social page. E.g browse to drive, then back to social.

ianopolous: So if I open social, then go to drive, then someone sends me a follow request, then I go back to social, it will show the new follow request? That's the test I care about from a UX perspective.

ianopolous: This is my reference: https://vuejs.org/guide/essentials/lifecycle.html#lifecycle-diagram

kevodwyer: > So if I open social, then go to drive, then someone sends me a follow request, then I go back to social, it will show the new follow request? That's the test I care about from a UX perspective.

Yes

ianopolous (approved):

PR #360, opened 2022-03-29
add markdown viewer

kevodwyer:
Also see Peergos PR: https://github.com/Peergos/Peergos/pull/912


src/views/Drive.vue
```diff
@@ -783,25 +822,33 @@ module.exports = {
...
-		    else if (app == "identity-proof")
-			this.showIdentityProof = true;
-		    else if (app == "hex")
-			this.showHexViewer = true;
-		    else if (app == "search")
-			this.showSearch = true;
-
-                    this.updateHistory(app, this.getPath, "");
+            let that = this;
+            Vue.nextTick(() => {
```
ianopolous, line 836: Can you explain the need for nextTick?

src/components/App.vue
```diff
@@ -77,6 +77,7 @@ const Social = require("../views/Social.vue");
 const Calendar = require("../views/Calendar.vue");
 const Chat = require("../views/Chat.vue");
 const Email = require("../views/Email.vue");
+const Markdown = require("../components/viewers/markdown/Markdown.vue");
```
ianopolous, line 80: Is this still needed?

src/views/Drive.vue
kevodwyer: I have added comments. Required due to side-effect of history navigation

src/components/App.vue
kevodwyer: unused code removed

PR #364, opened 2022-04-04
Make closing sub apps go back in history

ianopolous:


kevodwyer (approved):

kevodwyer:
I have done some testing with the new 'close' == back action. I am not convinced it is suitable for the md browser

ianopolous:
What issues did you hit?

kevodwyer:
> What issues did you hit?

If i got from one .md file to another .md file to another .md file i don't expect that clicking close will take me back to the previous .md file and so on just to close the md browser completely.

ianopolous:
Ah, you mean the close in the md browser, not the other apps?

ianopolous:
Yep, that's fair. I've reverted md-browser to going forward in history to drive.

kevodwyer (approved):

PR #373, opened 2022-04-12
folder download as a zip file

kevodwyer:
support downloading of a folder as a zip file
- only on browsers that support streaming
- limited to 4GiB
- zip contents not compressed

Also added a folder properties modal


ianopolous (approved):
Awesome. Stunning work!

PR #379, opened 2022-04-19
allow upload to a writable secret link

kevodwyer:


src/views/Drive.vue
```diff
@@ -1357,17 +1357,23 @@ module.exports = {
             let collectFuture = peergos.shared.util.Futures.incomplete();
             this.collectFiles(fromDnd, 0, filesToUpload, [], collectFuture);
             let that = this;
+            let urlProps = this.getPropsFromUrl();
+            if (urlProps.secretLink && !urlProps.writable) {
```
ianopolous, line 1361: would be good to avoid usign getPropsFromUrl
We already have a field for isSecretLink and can use currentDir.isWritable()

kevodwyer: done

ianopolous (approved):

PR #383, opened 2022-05-02
show toast with error message if upload cleanup fails

kevodwyer:
Reference: https://github.com/Peergos/Peergos/issues/932


ianopolous (approved):
Thanks

PR #386, opened 2022-05-04
handle potentially resuming a failed upload

kevodwyer:


src/views/Drive.vue
```diff
@@ -2064,7 +2092,8 @@ module.exports = {
 			let reader = new peergos.shared.user.fs.AsyncReader.ArrayBacked(empty);
 			this.currentDir.uploadFileJS(filename, reader, 0, 0,
 				false, false, that.getMirrorBatId(that.currentDir), this.context.network, this.context.crypto, function (len) { },
-				this.context.getTransactionService()
+				this.context.getTransactionService(),
+				f => peergos.shared.util.Futures.of(false)
```
ianopolous, line 2096: I think we're quite likely to want to resume if we're uploading a single large file?

kevodwyer: you are looking at the change to uploadEmptyFile(filename). The function bulkUpload() has the more complicated logic

ianopolous: So single file uploads are resumable?

kevodwyer: Yes

ianopolous: what about all the other calls to uploadFileJS() They don't have a resume parameter?

kevodwyer: Besides upload there are  other cases. 2 are uploading empty files. The third is creating a new todo file. I have just switched it to => true. Although it is very unlikely that would ever be necessary as todo cbor files are tiny.

ianopolous: Yeah I'm not worried about that. Just the case where I upload a single file through the ui.

PR #390, opened 2022-05-26
call new unblock function

kevodwyer:


ianopolous:
We also need to check if a user is blocked before we try and send a follow request. If they are then abort and tell the user.

kevodwyer:
> We also need to check if a user is blocked before we try and send a follow request. If they are then abort and tell the user.

done

PR #395, opened 2022-06-11
Refactor upload

kevodwyer:
remove recursive functions


ianopolous:
Were you able to reproduce the error?

ianopolous:
Does this affect upload and delete?

kevodwyer:
> Were you able to reproduce the error?

no

kevodwyer:
> Does this affect upload and delete?

affects upload only. Delete does not use recursion

ianopolous (approved):

PR #402, opened 2022-07-01
fix paste bugs

kevodwyer:
- can now paste via context menu into '/' folder
- removed redundant function sharedWithDataUpdate - it also had a bug that caused paste not to appear on '+' menu.
- removed redundant function showShareWithForFile (and calling functions) - it also had a latent bug

### DIFF
--- a/src/views/Drive.vue
+++ b/src/views/Drive.vue
@@ -701,8 +701,7 @@ module.exports = {
 				return false;
 			}
 
-			return (this.currentDir.isWritable() || this.currentDir.getFileProperties().name == '/')
-			    && target.isDirectory();
+			return target.isWritable() && target.isDirectory();
 		},
 	},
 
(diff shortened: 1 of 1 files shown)


ianopolous:
But no one can write to '/"?

src/views/Drive.vue
```diff
@@ -701,7 +701,8 @@ module.exports = {
 				return false;
 			}
 
-			return this.currentDir.isWritable() && target.isDirectory();
+			return (this.currentDir.isWritable() || this.currentDir.getFileProperties().name == '/')
+			    && target.isDirectory();
```
ianopolous, line 705: do we not need to check target is writable?

kevodwyer:
> But no one can write to '/"?

test is on the current folder. have changed test now

ianopolous (approved):

PR #403, opened 2022-07-05
chat api for Peergos Apps

kevodwyer:
- will add section to tech book after merge
- paired with peergos PR: https://github.com/Peergos/Peergos/pull/965
- example App will be uploaded to example-apps after merge.

API
The following HTTP actions are supported:

/peergos-api/v0/chat/
GET – Retrieve a list of all chats created by this App
Response code: 200 – success.
Response:
{chatId: chatId, title: title}

/peergos-api/v0/chat/:chatId
GET – Retrieve messages
Url Parameters:
from - paging from index
to - paging to index
Response code: 200 – success.
Response:
{messages:[], count: messagesRead}
Contents of messages array:
{type: 'Application', id: messageHash, text: text, author: author, timestamp: timestamp}
{type: 'Join', username: username, timestamp: timestamp}

POST - create chat
/peergos-api/v0/chat/
Request FormData
parameters: 
maxInvites
Response code: 201 – success.
Response header:
Location - chatId

/peergos-api/v0/chat/:chatId
PUT - send message
Request FormData
parameters: 
text - Contents of message
Response code: 201 – success.

### DIFF
--- a/assets/apps/sandbox/StreamSaver-sandbox.js
+++ b/assets/apps/sandbox/StreamSaver-sandbox.js
@@ -217,8 +217,8 @@
             // We never remove this iframes b/c it can interrupt saving
             makeIframe(evt.data.download)
           }
-        } else if (evt.data.filePath) {
-            fileDataCallback(evt.data.filePath, evt.data.requestId, evt.data.apiMethod, evt.data.bytes,
+        } else if (evt.data.filePath != null) {
+            fileDataCallback(evt.data.filePath, evt.data.requestId, evt.data.api, evt.data.apiMethod, evt.data.bytes,
                 evt.data.hasFormData, evt.data.params, evt.data.isFromRedirect);
         } else if (evt.data.seekLength) {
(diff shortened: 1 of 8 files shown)


jonaswre:
Can I listen to a event when a message is send or will it need to poll it?

ianopolous (approved):

jonaswre:
@kevodwyer messages will need to be polled right there is no reactive mechanism to recieve mesages?

kevodwyer:
yes, polling is currently required

jonaswre:
Thanks for the answer.

Have you ever considered a way to provide realtime communication? I know this would likely mean to provide a chat API through websockets. Which is not trivial, but that would be great as a feature on the roadmap.

ianopolous:
Yes, in the future we hope to enable opening a direct (p2p) stream (with permission) to another user running an app. This requires a lot more work behind the scenes though so it won't happen any time  soon.

PR #405, opened 2022-07-28
add Custom App support for HTTP HEAD and OPTIONS

kevodwyer:
support HEAD
support OPTIONS (minimal changes necessary by TiddlyWiki)
add new APP PERMISSION - description displayed 'Allow app to modify its own code'


ianopolous:
I think because permissions are user facing they should have simple names and explanations understandable by normal people.

kevodwyer:
> I think because permissions are user facing they should have simple names and explanations understandable by normal people.

Have updated description

ianopolous (approved):

PR #430, opened 2022-11-11
add retry to notablock request

kevodwyer:


src/components/App.vue
```diff
@@ -295,30 +295,39 @@ module.exports = {
...
-
 	checkIfDomainNeedsUnblocking() {
 	    if (this.network == null) return;
 	    var that = this;
 	    this.network.otherDomain().thenApply(function (domainOpt) {
-		if (domainOpt.isPresent()) {
+            if (domainOpt.isPresent()) {
+                let domain = domainOpt.get();
+                that.sendNotABlockRequest(domain, (errCb1) => {
+                    that.sendNotABlockRequest(domain, (errCb2) => {
```
ianopolous, line 305: We should only retry if a http 503 was returned.

kevodwyer: have added code to detect. needs to be deployed to test env to confirm behaviour

ianopolous: Can we also use recursion rather than 3 nested calls? Just have an extra parameter remainingAttempts?

PR #433, opened 2022-11-16
Make code editor work offline (read only)

ianopolous:


ianopolous:
@kevodwyer Do we need the test.js files? Maybe we can just delete them?

assets/apps/code-editor/sw.js
```diff
@@ -0,0 +1,327 @@
...
+self.addEventListener('activate', event => {
+    clients.claim();
+});
+
+self.onfetch = event => {
+    const url = event.request.url
+
+    let requestURL = new URL(url);
+    if (event.request.mode === 'navigate' && !requestURL.pathname.startsWith('/api')) {
+        if (requestURL.pathname == '/') {
```
kevodwyer, line 312: I don't think handling '/' is required for the code-editor

kevodwyer (approved):

kevodwyer (approved):

PR #439, opened 2022-11-23
fix refreshing htmlviewer/deep links to it

ianopolous:
@kevodwyer Do we want sandbox in this list as well?


kevodwyer:
> @kevodwyer Do we want sandbox in this list as well?

That will require more thought

kevodwyer (approved):

PR #440, opened 2022-11-23
improve htmlviewer app navigation handling

kevodwyer:


src/components/sandbox/AppSandbox.vue
```diff
@@ -809,33 +809,30 @@ module.exports = {
...
-                                if (!that.running) {
-                                    that.running = true;
-                                    that.startTitleDetection();
-                                    that.readFileOrFolder(headerFunc, path, params);
-                                } else {
+                                let fullPath = that.expandFilePath(path, isFromRedirect);
+                                let app = that.getApp(file, fullPath);
+                                let navigationPath = fullPath.substring(0, fullPath.lastIndexOf('/'));
+                                let navigationFilename = fullPath.substring(fullPath.lastIndexOf('/') + 1);
+                                if (app == 'htmlviewer' || app == 'editor' && fullPath.toLowerCase().endsWith('.html')) {
```
ianopolous, line 822: Why do we need the .html special case?

kevodwyer: because in router/index.js function getApp(file, path, writable)
if a html file is writable, getApp will return editor rather than htmlviewer

kevodwyer: actually i think i can get away with not having the special case

kevodwyer: nope. I remember now. Sometimes a html file is not recognised as mimetype html file, but just a text file (mimetype 'text/'. the above logic handles that case.

ianopolous: Surely we should fix the mimetype detection then?
Is that some extra characters outside printable ascii?

kevodwyer: what about the case where i just create a new text file in peergos itself and just have a .html extension?

kevodwyer: > what about the case where i just create a new text file in peergos itself and just have a .html extension?

just tested - that seems fine

kevodwyer: have removed special case. Next time it happens we can fix the mimetype detection.

src/components/sandbox/AppSandbox.vue
```diff
@@ -809,33 +809,30 @@ module.exports = {
...
-                                if (!that.running) {
-                                    that.running = true;
-                                    that.startTitleDetection();
-                                    that.readFileOrFolder(headerFunc, path, params);
-                                } else {
+                                let fullPath = that.expandFilePath(path, isFromRedirect);
+                                let app = that.getApp(file, fullPath);
+                                let navigationPath = fullPath.substring(0, fullPath.lastIndexOf('/'));
+                                let navigationFilename = fullPath.substring(fullPath.lastIndexOf('/') + 1);
+                                if (app == 'htmlviewer') {
```
ianopolous, line 822: when do we hit this case?

ianopolous (approved):

PR #450, opened 2022-11-29
bulk upload progress bar

kevodwyer:


src/views/Drive.vue
```diff
@@ -1535,36 +1540,38 @@ module.exports = {
                         //resetting .value tricks browser into allowing subsequent upload of same file(s)
                         document.getElementById('uploadFileInput').value = "";
                         document.getElementById('uploadDirectoriesInput').value = "";
-                        let progressBars = [];
+                        let name = 'bulkUpload-' + this.uuid();
+                        let title = "Encrypting and uploading file(s)";
```
ianopolous, line 1544: Can we say how many files we are uploading in title?

kevodwyer: we could but that may be misleading. The figure would be shown before accounting for files that already exist in the destination. A user may decide not to copy files that already exist via the confirmation prompt

ianopolous (approved):

PR #456, opened 2022-12-19
Redo app install. Numerous bugs squashed.

kevodwyer:
- Copying assets now uses FileWrapper.CopyTo
- Can now check for Apps that have updates on Launcher page
- Can update Apps from Launcher page


ianopolous (approved):
Awesome!

PR #458, opened 2022-12-24
code-editor: remove non-existent test files from sw.js cache 

kevodwyer:


ianopolous (approved):

ianopolous:
Were these what was stopping the service worker from installing?

kevodwyer:
> Were these what was stopping the service worker from installing?

No. SW worked but errors was shown in browser console (cache.addAll failing) and server logs (file not found)

PR #459, opened 2022-12-24
streamline app install

kevodwyer:
install app direct from app-gallery
install app direct from newsfeed

### DIFF
--- a/src/components/sandbox/AppSandbox.vue
+++ b/src/components/sandbox/AppSandbox.vue
@@ -20,6 +20,13 @@
             :consumer_func="prompt_consumer_func"
             :action="prompt_action"
         />
+            <AppInstall
+                v-if="showAppInstallation"
+                v-on:hide-app-installation="closeAppInstallation"
+                v-on:app-install-success="appInstallSuccess"
+                :appPropsFile="appInstallPropsFile"
+                :installFolder="appInstallFolder">
+            </AppInstall>
             <div class="modal-header" style="padding:0;min-height: 52px;">
                 <center><h2>{{ getFullPathForDisplay() }}</h2></center>
               <span style="position:absolute;top:0;right:0.2em;">
@@ -37,6 +44,7 @@
 <script>
 
 const AddToChat = require("AddToChat.vue");
+const AppInstall = require("AppInstall.vue");
 const AppPrompt = require("../prompt/AppPrompt.vue");
 
 const downloaderMixins = require("../../mixins/downloader/index.js");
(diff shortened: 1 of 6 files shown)


src/components/sandbox/AppSandbox.vue
```diff
@@ -632,6 +665,30 @@ module.exports = {
...
             return true;
         },
+        handleInstallAppRequest: function(header, path, apiMethod, data, hasFormData, params) {
+            let that = this;
+            if (!this.isAppGalleryMode || !hasFormData || apiMethod != 'POST') {
+                console.log('Install App API call is invalid');
+                that.buildResponse(headerFunc(), null, that.ACTION_FAILED);
+                return;
+            }
+            let requestBody = JSON.parse(new TextDecoder().decode(data));
```
ianopolous, line 675: This looks like an app can install any other app?

kevodwyer: functionality only available if isAppGalleryMode is set.
I have added a further restriction. Now using a normally invalid app name to trigger mode to guard against a malicious user.

ianopolous: Can you explain the security assumptions and model here and how it's enforced regarding installing an app?

ianopolous: Maybe "able to install apps" should itself be a permission?

kevodwyer: The Appsandbox is responsible for ensuring an app has api access via the permissions
The htmlviewer is an internal app that has no permissions and therefore no access to any api
The install-app api call is only relevant when the recommended apps button on the launcher page is pressed.
The recommended apps page is launched in the htmlviewer via the call to launchApp on line 467 of Launcher.vue
The App Sandbox recognises the app-gallery token and sets isAppGalleryMode=true.
It is not possible via construction for any other app to set isAppGalleryMode

ianopolous: Does it rely on $ not being allowing in app name? If so, where do we enforce that?

kevodwyer: Yes. we restrict the characters allowed in an app name. we enforce it in mixins/sandbox/index.js validateDisplayName(). this is called from verifyJSONFile() which is in turn called by AppInstall.vue when installing an app.

ianopolous: In the future I'd like to let anyone write their own "app store" which could request the "install apps" permission. Does any of this block that or make it harder? I'd like to keep that open.

kevodwyer: Yes, that path is available down the track.

ianopolous (approved):

PR #467, opened 2023-02-16
fix sandbox regressions

kevodwyer:
- handle path when app is launchable
- re-calculate thumbnail only if file is writable
- better handling of read only access

### DIFF
--- a/assets/apps/sandbox/sandbox.js
+++ b/assets/apps/sandbox/sandbox.js
@@ -72,8 +72,8 @@ function load(appName, appPath, allowBrowsing, theme, chatId, username, props) {
             path = path.length > 0 ? path + '&theme=' + theme : '?theme=' + theme;
             path = chatId.length > 0 ? path + '&chatId=' + chatId : path;
             path = path + '&username=' + username;
-            if (props.isFileWritable == true) {
-                path = path + '&isFileWritable=' + props.isFileWritable;
+            if (props.isPathWritable == true) {
+                path = path + '&isPathWritable=' + props.isPathWritable;
             }
             let src = allowBrowsing ? appPath.substring(1) : "index.html" + path;
             iframe.src= src;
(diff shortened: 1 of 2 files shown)


assets/apps/sandbox/sandbox.js
```diff
@@ -72,6 +72,9 @@ function load(appName, appPath, allowBrowsing, theme, chatId, username, props) {
             path = path.length > 0 ? path + '&theme=' + theme : '?theme=' + theme;
             path = chatId.length > 0 ? path + '&chatId=' + chatId : path;
             path = path + '&username=' + username;
+            if (props.isFileWritable == true) {
+                path = path + '&isFileWritable=' + props.isFileWritable;
```
ianopolous, line 76: What if it opens a dir? Can we keep it general by calling it isPathWritable?

kevodwyer: isFileWritable is specific. This property will not be set if app is a folder action. We currently do not have any apps that are folder actions (beyond my testing apps), i don't want to over-design until i have  a use case to confirm with.

ianopolous: Sure but if people write apps based on this and we change it they will break. It is talking about the path supplied being writable correct?

kevodwyer: It is only a convienence so a file app can conditionally show a save button.
I have no problem adding a isPathWritable for folders down the line, but i think that will require more thought, ie multiple folders.
isFileWritable has a specific meaning for a specific case.

ianopolous: What is already set in stone is that we have a path arg that is provided to the app right? All we are doing is providing metadata about that path. It's doesn't matter if it's a file or folder.

kevodwyer: I will change isFileWritable to isPathWritable

ianopolous (approved):

PR #473, opened 2023-03-21
opfs firefox support

kevodwyer:


ianopolous (approved):

vendor/priors/gwt.js
```diff
@@ -373,11 +373,16 @@ function getFileHandleCreateIfNecessary(filename, directory) {
         .then(blockDirHandle =>
             blockDirHandle.getFileHandle(filename, { create: true }));
 }
+let pendingWrites = new Map();
```
ianopolous, line 376: Clever! This is better than what I had in mind.

PR #478, opened 2023-03-23
disable OPFS for mobile browsers

kevodwyer:


vendor/priors/gwt.js
```diff
@@ -500,7 +501,8 @@ function isOPFSAvailable() {
     		    (function (p) {
     			return p.toString() === "[object SafariRemoteNotification]";
     		    })(!window["safari"] || safari.pushNotification);
-    if (!isSafari) {
+    let isMobile = /Mobi|Android/i.test(navigator.userAgent); // https://stackoverflow.com/a/24600597
```
ianopolous, line 504: it might be fine in iOS safari. Are we able to try it there?

kevodwyer: the SO link have a link to a mozilla article: https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent
in short - it work for safari

ianopolous: I mean opfs might be fine in iOS safari. We should test it.

ianopolous (approved):

PR #483, opened 2023-04-02
print preview is now available from an API call

kevodwyer:
this gets around the problem that was causing the call to window.print() failing


src/views/Drive.vue
```diff
@@ -2375,8 +2375,8 @@ module.exports = {
                 return peergos.shared.user.JavaScriptPoster.emptyArray();
         	} else {
                 var canvas = document.createElement('canvas');
-                canvas.width = window.innerWidth - 100;
-                canvas.height = window.innerHeight - 100;
+                canvas.width = 800;
```
ianopolous, line 2378: what are these numbers? how does this look on mobile?

ianopolous:
This looks super risky to me. Is this creating an unsandboxed iframe on the app subdomain and letting the app set html for it? What if it tries to include an external URL to exfiltrate data?

src/views/Drive.vue
kevodwyer: previously a blank image was created using the size of the screen on whatever device was being used. Using a standard size seems more apporiate. Image editor can resize if desired.

kevodwyer:
> This looks super risky to me. Is this creating an unsandboxed iframe on the app subdomain and letting the app set html for it? What if it tries to include an external URL to exfiltrate data?

Have added DOMpurify hooks to detect leaks as per documentation

assets/apps/sandbox/sw-sandbox.js
```diff
@@ -226,11 +230,10 @@ const oneMegBlockSize = 1024 * 1024 * 1;
 let defaultSrcCSP = "default-src 'self'; ";
 let scriptSrcCSP = "script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval'; ";
 let scriptSrcWithUnsafeCSP = "script-src 'self' 'unsafe-inline' 'unsafe-eval' 'wasm-unsafe-eval'; ";
-let sandboxCSP = "sandbox allow-same-origin allow-scripts allow-forms allow-modals;";
 
 let remainderCSP = "style-src 'self' 'unsafe-inline'; style-src-elem 'self' 'unsafe-inline' data:; font-src 'self' data:;img-src 'self' data: blob:;connect-src 'self' data:; media-src 'self' data:;";
-let defaultCSP = defaultSrcCSP + scriptSrcCSP + sandboxCSP + remainderCSP;
-let cspWithUnsafeEval = defaultSrcCSP + scriptSrcWithUnsafeCSP + sandboxCSP + remainderCSP;
+let defaultCSP = defaultSrcCSP + scriptSrcCSP + remainderCSP;
```
ianopolous, line 235: Does this mean the iframe is not sandboxed anymore? Was that the case until the print stuff anyway?

kevodwyer: this does not do anything. I added when i thought it was possible to do print preview from inside the app itself.

PR #485, opened 2023-04-04
Add packaging into container image

gsaslis:
This adds a first `Containerfile` that can be used to package Peergos into a container image. 

Notes: 

- a multi-stage build is used so that the final container image only contains the JRE + the app (no unnecessary JDK stuff)
- a `Dockerfile` is added as a symbolic link to the vendor-agnostic `Containerfile`, for convenience
- here's a link about "Containerfile vs. Dockerfile" - https://github.com/containers/buildah/discussions/3170
- this PR is still WIP - we should probably still add some github action that builds the container image, runs some smoke test against it and publishes to a container image registry.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,47 @@
+name: Create and publish a Container image
+
+on: [push]
+# TODO: revert to building on `main` branch only
+#on:
+#  push:
+#    branches: ['main']
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


ianopolous (approved):

ianopolous:
I can't see a way to make it run checks here, so I'll just merge it I think.

gsaslis:
@ianopolous I'd rather we figured out how to run them on the PR before merging to verify it works.

ianopolous:
ok I'll look into it

gsaslis:
@ianopolous how about this: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#approving-workflow-runs-on-a-pull-request-from-a-public-fork

PR #490, opened 2023-04-21
display file size in human readable form

kevodwyer:


ianopolous (commented):
I think unfortunately we have to use the non binary version here, as most people won't understand KiB.

kevodwyer:
> I think unfortunately we have to use the non binary version here, as most people won't understand KiB.

We already use KiB, MiB etc everywhere. Screenshot shows folder size and account size. 

Screenshot 2023-04-21 at 16 09 24

ianopolous:
Closes https://github.com/Peergos/Peergos/issues/1021

PR #500, opened 2023-05-05
new gallery

kevodwyer:
media now takes up full screen
code is much more concise
doesn't have the bug that afflicts the existing gallery whereby mixed media items results in media not loading


ianopolous:
Super exciting!

ianopolous:
Navigation with the arrow keys seems inconsistent on localhost. It sometimes jumps 2-5instead of 1.

ianopolous (approved):
AWESOME

PR #505, opened 2023-06-01
update vue 1 components to vue 2

kevodwyer:
Also delete old version of components already migrated, like password utils.


ianopolous (commented):
It seems like the passwordutil stuff has been deleted? Where are they now?

ianopolous (approved):

PR #512, opened 2023-07-25
Enable OPFS for Safari

kevodwyer:


vendor/priors/gwt.js
```diff
@@ -526,6 +518,7 @@ var batStoreCache;
 var accountStoreCache;
 var pkiStoreCache;
 var rootKeyCache;
+let SAFARI_CACHE_SIZE = 1024 * 1024 * 200;
```
ianopolous, line 521: Is this the default safari gives you? If you try to exceed it does it ask the user for permission?

ianopolous: I would suggest setting this to whatever safari gives you by default without warning.

PR #534, opened 2023-12-12
add latex support to builtin markdown viewer. also added to tui.markdown app

kevodwyer:
example

$$latex
\documentclass{article}
\begin{document}

$
f(x) = \int_{-\infty}^\infty \hat f(\xi)\,e^{2 \pi i \xi x} \, d\xi
$
\end{document}
$$

### DIFF
--- a/assets/apps/markup-viewer/init.js
+++ b/assets/apps/markup-viewer/init.js
@@ -43,12 +43,44 @@ window.addEventListener('message', function (e) {
 	function getCurrentFolder() {
 		return "";
 	}
+    function processMarkdownWithLatex(text) {
+        text = "\n" + text +"\n"; //add \n to catch token at very start and end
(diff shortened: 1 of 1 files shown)


ianopolous:
It also supports inline latex right? Like

# Heading $\pi$

ianopolous:
The following should work:
# heading
 Some text
 
 
 More text $\pi$ test of lkatex.
 
 $$ \frac{e^2}{1-r^2} $$

```latex
# heading
 Some text
 
 
 More text $\pi$ test of lkatex.
 
 $$ \frac{e^2}{1-r^2} $$
```

ianopolous:
We need those examples to work as is. 

This should work with extended mathjax: https://www.fabriziomusacchio.com/blog/2021-08-10-How_to_use_LaTeX_in_Markdown/

PR #543, opened 2024-02-23
add mathJax support to markdown viewer

kevodwyer:


ianopolous:
This doesn't seem to work for inline latex?

PR #544, opened 2024-02-28
add i18n support to calendar

kevodwyer:
This is not the final word for translations. It handles messages and visible text.


assets/apps/calendar/init.js
```diff
@@ -60,6 +60,66 @@ navigator.serviceWorker.getRegistration('./').then(swReg => {
     window.parent.postMessage("sw-registration-failure", parentHost)
 })
 
+let supported = ["en-GB"];
+let enGB = {
```
ianopolous, line 64: Could we include this as a separate json file that is fetched so it's easier to add new languages?

ianopolous: sure, separate js file is also fine.

ianopolous (approved):

PR #550, opened 2024-03-16
update social call is async. Do not allow other actions to proceed until completed

kevodwyer:


src/views/Social.vue
```diff
@@ -324,65 +330,66 @@ module.exports = {
         acceptAndReciprocate(req) {
             var that = this;
             this.showSpinner = true;
-            this.context.sendReplyFollowRequest(req, true, true)
-                .thenApply(function(success) {
-		    that.showSpinner = false;
-		    that.$toast(that.translate("SOCIAL.RECIPROCATED"))
-		    that.updateSocial();
```
ianopolous, line 331: we don't update social here? won't that mean the new friend won't show up?

src/views/Drive.vue
```diff
@@ -1035,20 +1037,21 @@ module.exports = {
...
-				this.context.getPaymentProperties(false).thenApply(function (paymentProps) {
-					if (paymentProps.isPaid()) {
-						that.paymentProperties = paymentProps;
-					} else
-						that.context.getPendingSpaceRequests().thenApply(reqs => {
-							if (reqs.toArray([]).length > 0)
-								that.$store.commit('USER_ADMIN', true);
-						});
-				});
+				this.updateSocial(() => {
```
ianopolous, line 1040: what was the problem here?

ianopolous (approved):

PR #560, opened 2024-03-26
resize confirm modal

kevodwyer:
confirm modal now does not take up the whole width of the screen
2 addtional small fixes
- when a new file is created from clicking an app on the launcher page, the toast to indicate a new file has been created now is dismissed after a few seconds
- the markup viewer now waits a small amount of time before updating the link/image resources in the file.


ianopolous:
How does it look on mobile? There you probably want the full width?

ianopolous (approved):

PR #563, opened 2024-03-28
Add progress bar for multi-select move and copy actions.

kevodwyer:


src/views/Drive.vue
```diff
@@ -2339,9 +2351,29 @@ module.exports = {
             }
             this.closePasteMenu();
             that.showSpinner = true;
```
ianopolous, line 2353: Can we drop the spinner if we have a progress bar?

ianopolous: Why? We don't have a spinner during uploads.

PR #567, opened 2024-04-08
updateSocial() race condition causing CAS Exception

kevodwyer:
updateSocial called in postLogin and in setup(). The seond call would start before the first had finished. this leads to a CAS exception

### DIFF
--- a/src/components/Login.vue
+++ b/src/components/Login.vue
@@ -186,21 +186,19 @@ module.exports = {
 
             that.$emit("initApp")
 
-            that.updateSocial(() => {
-                // that.$store.commit('CURRENT_MODAL', 'ModalTour');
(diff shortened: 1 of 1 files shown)


src/components/Login.vue
```diff
@@ -186,21 +186,19 @@ module.exports = {
...
-                            console.log("Offline support enabled");
-                        } else {
-                            console.log("Offline support not available");
-                        }
-                    });
-                }
-            });
+            // that.$store.commit('CURRENT_MODAL', 'ModalTour');
+
+            console.log("Signing in/up took " + (Date.now()-creationStart)+" mS from function call");
```
ianopolous, line 191: where is update social called now?

ianopolous: Drive.setup

ianopolous (approved):

PR #595, opened 2024-06-18
remove inbuilt tasks app

kevodwyer:


assets/apps/sandbox/sw-sandbox.js
```diff
@@ -18,6 +18,7 @@ let filePickerRequest = apiRequest + "/file-picker/";
 let foldersRequest = apiRequest + "/folders/";
 let printRequest = apiRequest + "/print/";
 let profileRequest = apiRequest + "/profile/";
+let tasksRequest = apiRequest + "/tasks/";
```
ianopolous, line 21: why does this need a new api?

ianopolous: Can you explain this new api in the PR? I don't want to bake in some weird file format to the api.

ianopolous (approved):

PR #597, opened 2024-06-20
introduce method to ease installation of recommended app for file on Drive and Newsfeed

kevodwyer:
Now if a user double clicks on a file for which their is a recommended app:
The recomended app page is dispayed and the app in question is highlighted
- if app is subsequently installed, then it will be used to open the file
- if app is not installed, the default app ie 'hex' will be used.

anchor tags have been added to index.html for recommended-apps

### DIFF
--- a/src/views/NewsFeed.vue
+++ b/src/views/NewsFeed.vue
@@ -1153,6 +1153,7 @@ module.exports = {
             let path = "/peergos/recommended-apps/";
             this.context.getByPath(path + "index.html").thenApply(function(fileOpt){
                 if (fileOpt.ref != null && fileOpt.get().getFileProperties().sizeLow() > 20) {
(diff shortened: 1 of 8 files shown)


src/views/NewsFeed.vue
```diff
@@ -1103,8 +1130,13 @@ module.exports = {
                 let inbuiltApps = this.getInbuiltApps(file);
                 if (userApps.length == 0) {
                     if (inbuiltApps.length == 1) {
-                        if (inbuiltApps[0].name == 'hex') {
-                            this.openFileOrDir("Drive", path, {filename:""});
+                        if (inbuiltApps[0].name == 'hex' || inbuiltApps[0].name == 'editor') {
```
ianopolous, line 1133: Why is editor here?

ianopolous: Does that mean it's going to ask me to install an app every time I open a .java file?

ianopolous (approved):

PR #603, opened 2024-07-21
regression - calendar secret link

kevodwyer:


src/components/drive/DriveShare.vue
```diff
@@ -344,7 +344,7 @@ module.exports = {
 				id: "secret_link_" + name,
 				isFile: isFile,
 				shareFolderWithFile: shareFolderWithFile,
-                autoOpen: shareFolderWithFile === true ? true: false,
+                autoOpen: (shareFolderWithFile === true || !isFile) ? true: false,
```
ianopolous, line 347: What if I'm in drive and just want to share a folder (without opening anything)?

ianopolous (approved):

PR #606, opened 2024-07-24
Fix/secret link cleanup

ianopolous:
Simplify secret link modal

Remove unnecessary args from secret links

Fix hanging secret link operations (I think it is the second such operation failing because the underlying FileWrapper wasn't updated after the first operation)


ianopolous:
A java side fix resolves the errors I was seeing.

PR #609, opened 2024-08-26
re-create md/notes editor iframe when navigating

kevodwyer:
The fix was done as mathJAX was erroring when initialised more then once on the same page


ianopolous (commented):
I'd like to understand the consequences of  this. Do resizes not  work now? Is it slower to navigate to md links? Is there no way to catch the mathjax  error? What about not showing the not  found error?

PR #613, opened 2024-09-06
secret link modal styling

kevodwyer:
add a vertical scroll bar for small screens (mobile phones)


ianopolous:
What's all the Cambria Math about?

PR #614, opened 2024-09-10
add sandbox mailbox api

kevodwyer:
This is used by the email client app (was previously an internal app)


src/i18n/en-GB.js
```diff
@@ -469,6 +469,7 @@ module.exports = {
     "SANDBOX.EDIT.FILE":"Can modify file chosen by user",
     "SANDBOX.READ.FOLDER":"Can read selected files of the associated types from folder chosen by user",
     "SANDBOX.MESSAGES":"Can exchange messages with friends",
+    "SANDBOX.MAILBOX":"Can post and receive messages via a mailbox",
```
ianopolous, line 472: This is email right? Not just a generic message?

ianopolous (approved):

PR #615, opened 2024-09-16
add cross-origin-isolated to sandbox iframe

kevodwyer:
Using hint from:
https://issues.chromium.org/issues/346852676

And text from:
https://web.dev/articles/cross-origin-isolation-guide

Have added allow attribute to iframe and set it to ' cross-origin-isolated '

Have tested in chrome and inner most iframe now has window.crossOriginIsolated = true


ianopolous:
And it's still fine in firefox and safari?

PR #618, opened 2024-09-18
bugfix - multiple app launches

kevodwyer:
A bug was encountered where if peergos was loaded over a bandwidth limited network, an app would potentially be launched multiple times

### DIFF
--- a/src/views/Drive.vue
+++ b/src/views/Drive.vue
@@ -558,7 +558,7 @@ module.exports = {
             progressBarUpdateFrequency: 50,
             zipAndDownloadFoldersCount: 0,
             htmlAnchor: "",
-            previouslyOpenedApp: {filename: '', app: ''},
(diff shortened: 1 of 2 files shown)


src/views/Drive.vue
```diff
@@ -1260,38 +1263,40 @@ module.exports = {
         },
 	    openInApp(args, app) {
                 if (app == null || app == "" || app == "Drive") {
+                    this.previouslyOpenedApp = {filename: '', app: ''};
                     this.closeApps();
-                    return
+                    return;
+                }
+                if (this.previouslyOpenedApp.filename == args.filename && this.previouslyOpenedApp.app == app) {
```
ianopolous, line 1270: I would expect opening a file with the same name in a different dir to be common, e.g. index.html

ianopolous (approved):

PR #619, opened 2024-09-25
update graalvm libs to 24.1.0

kevodwyer:
this log message appears repeatedly when building.

[exec] For more information see: https://www.graalvm.org/latest/reference-manual/embed-languages/#runtime-optimization-support.
     [exec] To disable this warning use the '--engine.WarnInterpreterOnly=false' option or the '-Dpolyglot.engine.WarnInterpreterOnly=false' system property.
     [exec] [To redirect Truffle log output to a file use one of the following options:
     [exec] * '--log.file=<path>' if the option is passed using a guest language launcher.
     [exec] * '-Dpolyglot.log.file=<path>' if the option is passed using the host Java launcher.
     [exec] * Configure logging using the polyglot embedding API.]
     [exec] [engine] WARNING: The polyglot engine uses a fallback runtime that does not support runtime compilation to native code.
     [exec] Execution without runtime compilation will negatively impact the guest application performance.
     [exec] The following cause was found: JVMCI is required to enable optimizations. Pass -XX:+EnableJVMCI as a virtual machine argument to the java executable to resolve this.


ianopolous:
Can you add -XX:+EnableJVMCI

ianopolous (approved):

PR #631, opened 2024-10-28
App Manifest - check type of string fields

kevodwyer:


ianopolous (approved):
Is this for issue 14?

PR #643, opened 2024-11-18
when reading a block from OPFS use a read only AccessHandler

kevodwyer:


ianopolous (approved):
Does this make it faster?

PR #649, opened 2024-11-21
refresh folder from breadcrumb

kevodwyer:
as a side-effect this both fixes a lot of related stale/out-of-date scenarios and means a folder is populated progressively as actions occur


ianopolous (approved):

ianopolous:
Why doesn't .get() work?

ianopolous:
I'd like to understand why.

ianopolous:
Right so because it is Optional.empty(). Can not use the api, like isPresent() rather than GWT implementation details?

ianopolous (approved):
nice

PR #653, opened 2024-11-29
Drive - Add progress bar to copy and move actions

kevodwyer:


ianopolous:
Does this still show spinner and progress bar?

ianopolous (approved):

PR #657, opened 2024-12-12
Launcher - Optimise loading app icons

kevodwyer:
- Cache app icons
- Update shortcut table after sandbox closed


ianopolous (approved):
Nice! I have noticed how slow this is.

PR #668, opened 2025-02-04
do not cache blocks on localhost

ianopolous:


ianopolous:
What's the reasoning here?

PR #680, opened 2025-03-11
change service worker ping to use XMLHttpRequest rather than fetch()

kevodwyer:
This seems to stop the servie worker from shutting down

### DIFF
--- a/assets/apps/sandbox/sandbox.js
+++ b/assets/apps/sandbox/sandbox.js
@@ -74,20 +74,19 @@ function respondToLoadedChunk(bytes) {
     streamWriter.write(bytes);
 }
 function startPing(pingUrl) {
-    fetchRequest(pingUrl);
+    sendPingRequest(pingUrl);
     setTimeout(() => this.startPing(pingUrl), 5000);
 }
-function fetchRequest(url) {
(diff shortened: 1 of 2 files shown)


assets/apps/sandbox/sandbox.js
```diff
@@ -74,6 +74,19 @@ function respondToLoadedChunk(bytes) {
...
-    fetch(pingUrl);
+    sendPingRequest(pingUrl);
     setTimeout(() => this.startPing(pingUrl), 5000);
 }
+function sendPingRequest(url) {
+    var req = new XMLHttpRequest();
+    req.open('GET', url);
+    req.onload = function() {
+        if (!req.status == 200) {
+            console.log('sendPingRequest-status-!200')
```
ianopolous, line 85: Maybe log the actual status?

assets/apps/sandbox/sandbox.js
```diff
@@ -74,6 +74,19 @@ function respondToLoadedChunk(bytes) {
...
+        }
+    };
+    req.onerror = function(e) {
+        console.log('sendPingRequest-onerror')
```
ianopolous, line 89: and log the error here

PR #681, opened 2025-03-11
format file size on search results page

kevodwyer:


src/components/Search.vue
```diff
@@ -144,8 +144,8 @@ module.exports = {
     },
     methods: {
     convertBytesToHumanReadable:function(bytes) {
-        if (bytes == 0)
-            return "";
+        if (bytes == "")
```
ianopolous, line 147: why is it ""?

ianopolous (approved):

PR #687, opened 2025-03-17
fallback to localStorage for auto-login if indexedDB not available

kevodwyer:
This fixes stay logged in on Macs and iOS

### DIFF
--- a/vendor/priors/gwt.js
+++ b/vendor/priors/gwt.js
@@ -1152,7 +1152,7 @@ function getRootKeyEntryFromCacheProm() {
                 }
             });
         } else {
-            let val = window.localStorage.getItem('rootKey');
+            let val = window.localStorage.getItem('logindata-rootKey');
             if (val == null) {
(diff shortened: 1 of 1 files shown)


vendor/priors/gwt.js
```diff
@@ -1288,11 +1320,24 @@ function directGetEntryDataFromCacheProm(key) {
...
+    }
+    future.complete(convertToByteArray(data));
+}
 function clearAccountCacheFully(func) {
     if (accountStoreCache.isCachingEnabled) {
         clearIDBKV(accountStoreCache.cacheAccountStore).then(res => func());
     } else {
+        for (var i = 0; i < window.localStorage.length; i++) {
+            let key = window.localStorage.key(i);
+            if (key != 'rootKey') {
```
ianopolous, line 1337: Would it be safer to have a prefix for the key here? In case we use localstorage for something else?

ianopolous (approved):

PR #688, opened 2025-03-17
drive - improve upload progress bar

kevodwyer:


src/views/Drive.vue
```diff
@@ -1910,7 +1910,7 @@ module.exports = {
                 let isWritableSecretLink = this.isSecretLink && this.currentDir.isWritable();
                 let totalSize = 0;
                 for(var i=0; i < files.length; i++) {
-                    totalSize += (files[i].size + (4096 - (files[i].size % 4096)));
+                    totalSize += ((files[i].size * 2) + (4096 - (files[i].size % 4096)));
```
ianopolous, line 1913: I think we need to double the last bit too

ianopolous (approved):

PR #694, opened 2025-04-18
opfs - refactor reading data

kevodwyer:


vendor/priors/gwt.js
```diff
@@ -416,30 +422,47 @@ function setOPFSKV(filename, value, directory) {
     });
     return future;
 }
+function waitFor(pendingReadFuture, future) {
```
ianopolous, line 425: why do we need waitFor? Can't we just use the .thenApply fo the future or equivalent for promise?

ianopolous:
Did you test this with a 10k file dir?

vendor/priors/gwt.js
```diff
@@ -397,7 +402,9 @@ function getFileHandleCreateIfNecessary(filename, directory) {
             blockDirHandle.getFileHandle(filename, { create: true }));
 }
 let pendingWrites = new Map();
-
+let pendingReads = new Map();
```
ianopolous, line 405: Can you comment with what the types are here? If the values are lists, why can't be a single future?

ianopolous: If I have a pendingReads map<String, Future<byte[]>> then if n other reads come in can't I just call map.get(k).thenComplete(ourFuture) which would complete all waiting futures as soon as the first one is done?
So no need for a list? Maybe I'm being dense.

PR #707, opened 2025-06-09
fix spinner position on search page

kevodwyer:


ianopolous:
Does this mean we can click results before completion?

PR #714, opened 2025-06-26
Sandbox app improvements

kevodwyer:
Fix regression with app-dev-mode not working
Template apps like forum can now manage membership from app context menu
Fix bugs on Group.vue
Fix minor bugs when installing app from NewsFeed and with file upload within app

### DIFF
--- a/src/views/Launcher.vue
+++ b/src/views/Launcher.vue
@@ -3,7 +3,7 @@
 	<AppHeader>
 	</AppHeader>
         <div class="modal-body">
-            <Spinner v-if="showSpinner"></Spinner>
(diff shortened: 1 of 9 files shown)


src/views/Launcher.vue
```diff
@@ -466,6 +495,187 @@ module.exports = {
...
+                            that.showErrorMessage("Unable to add members to chat");
+                            console.log(err);
+                            future.complete(false);
+                        });
+                    });
+                });
+            }
+            return future;
+        },
+        getPublicKeyHashes: function(usernames) {
```
ianopolous, line 548: bear in mind that usernames will go away when we remove the PKI requirement.

PR #752, opened 2026-04-21
Hallo iedereen! Dutch localisation added

Harry2L8:
I decided to follow the naming scheme of the English reference file, though there is no difference in spelling between Belgian Dutch and Dutch Dutch in this file.


ianopolous (approved):
Awesome! Thank you

PR #756, opened 2026-07-15
Fix empty folders being silently dropped when uploading a folder

arisoik:
Hello Peergos team,
During usage of Desktop Peergos app I found the following issue. Since I'm not a Vue.js developer I used Claude Code to fix and document it. I tested it in Firefox and Chromium browsers to see the expected behaviors after the fix.

## Bug

Uploading a folder tree that contains an empty subfolder (or a
subfolder whose only content is a filtered-out `.DS_Store`) silently
drops that subfolder - it never appears at the destination, with no
error or warning. Reported via drive folder upload testing: a folder
with 0 direct files and 8 empty sub-folders vanished entirely.

## Root cause

Folder creation is inferred entirely from the paths of uploaded
files (`processFileUpload` → `uploadFileJS`, `src/views/Drive.vue`).
A folder only gets added to the list of folders to create on the
server if at least one file lives in it somewhere. A folder with no
files anywhere in its own subtree contributes nothing, so it's never
created - this affects both the drag-and-drop upload path and the
"Upload folder" menu button.

For the "Upload folder" button specifically, this is a hard
platform limitation, not a logic bug: it uses
`<input type="file" webkitdirectory>`, and that API's `FileList`
**only ever contains files** - the browser gives JavaScript no
information about empty directories at all, so there's nothing in
the existing code path to fix.

## Fix (commit 1)

- **Drag-and-drop**: `getEntries()` already walks real directory
  entries via `webkitGetAsEntry()`/`readEntries()`, so it now records
  every directory path it visits, not just ones containing files.
  `processFileUpload()` takes a new `extraDirectories` param and
  pre-seeds the upload's folder list with any of those paths that no
  file already covers, so they still get created with zero files.
- **"Upload folder" button**: rebuilt to prefer the File System
  Access API (`showDirectoryPicker()`) where available, which *can*
  walk real directory handles including empty ones. Falls back to
  the original `<input webkitdirectory>` behavior where the API
  isn't supported.

### Platform/browser coverage

| Environment | Engine | Result |
|---|---|---|
| Chrome / Edge | Chromium | Both button and drag-and-drop fixed |
| Firefox | Gecko | Drag-and-drop fixed; button falls back (see commit 2) |
| Safari | WebKit | Same as Firefox |
| Windows desktop app | WebView2 (Chromium) | Should work like Chrome |
| macOS desktop app | WKWebView (WebKit) | Same as Firefox/Safari - confirmed via the native Swift delegate code |
| Linux desktop app | none - runs as a plain server, opens in the user's own browser | Depends on browser, as above |
| Android app | N/A | Unaffected - has its own native directory-picking path (`Android.notifyDirectoryRequest()`) that this change doesn't touch |

## User-facing notice (commit 2)

Where the fallback is used, silently reproducing the old limitation
felt worse than saying nothing, so an info toast now explains it and
points at drag-and-drop as the workaround. Translated into all
supported GUI languages (machine/model-quality translations for the
8 non-English locales - worth a native-speaker pass before/after
merge, especially Korean and Chinese).

## Testing

Manually verified by the reporter:
- Drag-and-drop with an empty subfolder: **works**
- "Upload folder" button on Firefox: falls back correctly, toast
  appears
- Chromium-based desktop apps: not yet tested live, reasoned about
  from the wrapper source (WebView2 is Chromium) but not confirmed
  hands-on

Not yet tested: Safari, macOS desktop app, Android (expected
unaffected), very large/deep folder trees.

### DIFF
--- a/src/components/drive/DriveHeader.vue
+++ b/src/components/drive/DriveHeader.vue
@@ -139,6 +139,7 @@ module.exports = {
                                 if (e && e.name !== 'AbortError') console.log(e);
                             });
                         } else {
+                            this.$toast.info(this.translate("DRIVE.UPLOAD.FOLDER.EMPTY.WARNING"), {id: 'upload-folder-limitation'});
                             document.getElementById('uploadDirectoriesInput').click();
                         }
 		},
--- a/src/components/drive/DriveHeader.vue
+++ b/src/components/drive/DriveHeader.vue
@@ -127,9 +127,44 @@ module.exports = {
                         if (typeof Android !== 'undefined') {
                             Android.notifyDirectoryRequest();
                             document.getElementById('uploadFileInput').click();
+                        } else if (typeof window.showDirectoryPicker === 'function') {
+                            // Unlike <input webkitdirectory>, this can see empty folders too.
+                            let that = this;
+                            window.showDirectoryPicker().then(function(dirHandle) {
+                                let files = [], directories = [];
+                                that.walkDirectoryHandle(dirHandle, '/' + dirHandle.name, files, directories).then(function() {
+                                    that.$emit('filesToUpload', {files: files, directories: directories});
+                                });
+                            }).catch(function(e) {
+                                if (e && e.name !== 'AbortError') console.log(e);
+                            });
                         } else {
                             document.getElementById('uploadDirectoriesInput').click();
                         }
+		},
+
+		// Walks a FileSystemDirectoryHandle, collecting files (tagged with their folder's
+		// path) and every directory path seen, including empty ones.
+		walkDirectoryHandle(dirHandle, path, files, directories) {
+			let that = this;
+			directories.push(path);
+			let iterator = dirHandle.entries();
+			function readNext() {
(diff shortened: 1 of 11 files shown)


ianopolous:
What is the use case for wanting the empty folders created too?

arisoik:
To be honest I don't have a specific use case in mind but I would want such behavior for exact sync of source and destination paths. If it was a design choice I'm ok but I believe that at least user should know somehow this kind of behavior.

ianopolous (approved):
Thank you. I will wait to merge until the pending drive mount is merged.

ianopolous:
Can we make it only show the toast if such a dir is encountered?

arisoik:
I did now a quick research what other tools are doing and in most cases they are skipping empty folders or add a flag to include them. I'm sorry for not doing this before but it is ok for me to decline this PR if you want because you are doing what most users seems to expect.

ianopolous:
Yep that's fine. For what it is worth, our sync client does create empty folders.

PR #767, opened 2026-08-16
Redesign the sync view

arisoik:
Rewrites the sync view around a single summary plus one row per folder.

## Summary

Four states only — up to date, syncing, needs attention, paused — so the headline is
always one of a small, predictable set. Counts pluralise, and the banner never claims
everything is settled while a folder still has work outstanding.

## Folders

Each row shows this device and the Drive folder either side of a direction arrow,
with a status pill, the current activity, and the delete behaviour of the pair. Long
paths clamp and expand on tap rather than pushing the layout around, and the status
column is fixed so the remote paths line up down the list.

A folder the pass has not reached yet reads "Waiting to sync" instead of showing its
previous result. The pass takes folders in order, one at a time, so this is derived
from the status payload rather than guessed: anything after the folder currently
syncing is still queued.

Pause and Resume sit next to Sync now. Pause needs the server endpoint from the
Peergos PR; without it that one button reports an error and the rest of the view is
unaffected.

## Details

- Relative timestamps once a folder has been quiet for a minute ("5 min ago"); nothing
  is shown while the cycle is healthy
- Mobile data toggle per pair, with a toast confirming when it takes effect
- Responsive down to narrow screens, 48px touch targets, reduced-motion honoured
- 22 aria attributes across the interactive parts
- 41 new i18n keys, 2 reworded; other locales fall back to English as usual

## Testing

A headless suite drives the real component through recorded server payloads: 33
checks covering the state machine, pause and resume, removal, queued folders,
pluralisation and the summary text. Also replayed live sessions against a running
local server with two sync pairs, and checked layout at desktop and narrow widths
with the longest translations of each label.


ianopolous (approved):
Wonderful! Much better looking.

PR #782, opened 2026-09-16
Redesign the drive and the views around it, and fix what the browser tests found underneath

arisoik:
The drive and the views around it are rebuilt on the status card surfaces: spinners and the
load bar that wait before appearing rather than flickering, progress measured by files rather
than bytes, menus that toggle from their own button, sticky headers, and a table that drops the
rules it was drawing twice. A phone's collapsed sidebar no longer follows you back to the
desktop, and the launcher and Shared With empty states are translated into every language.

Building that turned up four bugs underneath it that have nothing to do with the redesign. They
are in here because the browser tests found them and they are worth fixing on their own:

**Uploads failed silently.** A folder moved or deleted between choosing the files and starting
the upload left the progress bar ticking over an upload that would never happen, behind a future
that never settled. It now says so and stops. A link that cannot be opened reports instead of
doing nothing.

**The drive could stay behind its spinner at first sign-in.** Listing a folder raced the launcher
app's own initialisation, and the loser came back with a CAS conflict that nothing caught.
Listing a folder needs no launcher app, so it no longer waits for one.

**The calendar could come up empty and stay that way.** `updateCalendarList` guarded the
shared-calendar branch against a failed listing but not the branch for your own calendar, so a
single failed read left the count short, the future unresolved, and the load waiting on it never
ran. The properties write had the same gap. Both now settle and load anyway.

**The calendar could lose its calendars entirely.** The host sends the whole list in one `load`
message, gated on a handshake flag that is set once by whichever document answered a ping and is
never cleared. When the frame's document is replaced, that message is posted into a page that has
no listener yet, and `postMessage` keeps nothing for a listener that registers later — so it is
gone, with no acknowledgement and no retry. The app is left with an empty grid, no calendar to
save into, and no spinner to say why. The app now announces itself in every document it loads,
and the host re-sends the load to a frame that arrived after one went out.

The browser tests also survive a slow runner better: a navigation that times out but landed is no
longer fatal, the waits on a write get the budget the rest of the suite uses, and a failure prints
what the app and the host actually hold instead of a bare timeout.

---

**On testing the last one.** That test passes in CI some of the time — it failed two runs and
passed another on unchanged code — so a green run does not confirm the fix. It reproduces under a
constrained CPU, which is what CI runners have and a workstation does not:

    taskset -c 0,1 java -cp ../server/Peergos.jar Suite.java firefox

On two cores the whole-share test failed 24 of 56 rounds before the fix and 0 of 14 after it.

**Two related hazards are left alone deliberately**, both pre-existing: `sendPing` reaches through
`iframe.contentWindow` with no null check while `postMessage` schedules its retry *after* calling
it, so a throw there loses a message for good; and a frame whose `hello` arrives before the host
has installed its listener is still ignored. I tried the first and it made failures measurably
worse in combination, so it is not in here.

Verified: full browser suite passes on firefox and chromium. No horizontal scroll and nothing
escaping the viewport across Drive, Shared, Social and Calendar at 390, 768 and 1280px on
chromium and at 500, 768 and 1280px on firefox, which will not size a headless window narrower
than 500. A 120-character filename truncates with an ellipsis rather than pushing the table
sideways. Not checked: the app shell's web view, keyboard focus order, 200% zoom, tap-target
sizes.


src/components/drive/DriveTable.vue
```diff
@@ -1,253 +1,472 @@
...
+			</tr>
+		</tbody>
+	</table>
+</template>
+
+<script>
+const AppButton = require("../AppButton.vue");
+const AppIcon = require("../AppIcon.vue");
+const ShareMark = require("./ShareMark.vue");
+const mixins = require("../../mixins/downloader/index.js");
```
ianopolous: can you rename this to downloader

ianopolous:
Awesome!

